PR #68, opened 2015-03-04
completed the linking of the shared library

HannesSommer:
Without this the shared library on will be incomplete (have unresolved symbols). This is fine _if_ one correctly (potentially including order!) provides all the missing shared libraries when linking a final executable, which needs to be complete.

This fails for example because the build system is messing up the order of the shared libraries dependencies. For example catkin will produce in unlucky situations a wrong sequence of the shared libraries resulting in a linking error (e.g. undefined libnabo symbols) for the final executable.

Btw: The order constraint for the gnu linker ```ld``` (when invoked with ```--as-needed```, which is default for ```g++``` or ```gcc```) is to have libnabo at least once behind the first libpointmatcher in the list of shared libraries.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -257,6 +257,7 @@ else(APPLE)
 
 	if (SHARED_LIBS)
 		add_library(pointmatcher SHARED ${POINTMATCHER_SRC} ${POINTMATCHER_HEADERS})
+		target_link_libraries(pointmatcher ${Boost_LIBRARIES} ${NABO_LIBRARY})
 		install(TARGETS pointmatcher LIBRARY DESTINATION ${INSTALL_LIB_DIR})
 	else(SHARED_LIBS)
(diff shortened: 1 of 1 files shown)


pomerlef:
It's reverted. I was doing something in parallel and thought about it more 30 min later...

HannesSommer:
Because of the failing of this revert ( http://jenkins.ros.org/job/devel-indigo-libpointmatcher/ARCH_PARAM=amd64,UBUNTU_PARAM=trusty,label=devel/121/changes), I've just looked at the history of the builds and apparently my change did not cause this problem (still redundant and reverting was right). 
But apparently there is an issue with libnabo's installation, at least for this environment (I guess cmakeConfig). Is that a problem? I could have a look.

pomerlef:
yes, that would be helpful if you could have a look at this. I getting overwhelmed by support those days...

HannesSommer:
I'm on it :). And I have a good idea already: libnabo has not been "released" (bloom) since an import fix from you, that most likely solves that. I'll try a more current libnabo release.

Issue #66, opened 2015-02-25
installing YAML-pm into /usr/local/lib/ directory...


drhalftone:
How do I install the YAML static library into the /usr/local/lib directory?  I can use sudo make install for NABO and PM, but what about YAML?

drhalftone:
Obviously, I can manually copy it, but is there a sudo make install like script to install it along with libpointmatcher.a?

pomerlef:
This library was packaged recently and it seems that there is not installation target...

I'll look into that in the next days.

Issue #70, opened 2015-03-07
yaml-cpp-pm isn't compiled anymore under some build conditions


HannesSommer:
I strongly suspect that this line 
```cmake
 get_property(yaml-cpp-pm_LIB TARGET yaml-cpp-pm PROPERTY LOCATION)
```
(https://github.com/ethz-asl/libpointmatcher/commit/a9b72eb323#diff-af3b638bc2a3e6c650974192a53c7291R190)

Isn't exactly what we want, because it resolves to a path and not to a target name. It seems not implicitly create a target dependency for ```pointmatcher``` anymore:
```
[ 37%] Building CXX object CMakeFiles/pointmatcher.dir/pointmatcher/TransformationCheckersImpl.cpp.o
[ 39%] Building CXX object CMakeFiles/pointmatcher.dir/pointmatcher/InspectorsImpl.cpp.o
make[2]: *** No rule to make target `contrib/yaml-cpp-pm/libyaml-cpp-pm.a', needed by `libpointmatcher.so.1.2.1'.  Stop.
```
(taken from http://jenkins.ros.org/job/prerelease-indigo-libpointmatcher/ARCH_PARAM=amd64,UBUNTU_PARAM=trusty,label=prerelease/1/console).

It can be reproduced with a ```make pointmatcher``` when ```yaml-cpp-pm``` has not been built or its output removed before. Apparently this is what is issued on jenkins.ros.org. (it works with ```make```, because ```yaml-cpp-pm``` is within the default targets and _before_ ```pointmatcher```.

The simple solution before a9b72eb323:

```cmake
set (EXTERNAL_LIBS ${EXTERNAL_LIBS} yaml-cpp-pm )
```
would make cmake to let ```libpointmatcher``` depend on ```yaml-cpp-pm``` (in the sense of cmake's target dependencies), in combination with 
```cmake
target_link_libraries(pointmatcher ${EXTERNAL_LIBS})
```
(not changed with a9b72eb323)

@pomerlef , what was the reason for this location property based solution? 

One can fix it with adding an explicit dependency. E.g. :
```cmake
add_dependencies(pointmatcher yaml-cpp-pm)
```

But this needs to come after the ```pointmatcher``` target's declaration. So maybe we should introduce another collecting variable for dependencies? E.g. ```POINTMATCHER_EXTRA_DEPS``` ?

HannesSommer:
As a suggested solution I've just created #71 .

pomerlef:
[...]

PR #71, opened 2015-03-07
addressing #70 by introducing EXTRA_DEPS and adding yam-cpp-pm if necessary

HannesSommer:
* introduced EXTRA_DEPS variable to collect targets the pointmatcher target should depend on
* added yam-cpp-pm to EXTRA_DEPS in case of NOT USE_SYSTEM_YAML_CPP

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,6 +94,7 @@ include(UseDoxygen)
 
 # initially
 set(EXTERNAL_LIBS "")
+set(EXTRA_DEPS "")
 
 # compile local version of gtest and yaml-cpp
 add_subdirectory(contrib)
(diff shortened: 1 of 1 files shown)


pomerlef:
I lost track of where we are with that merge request. Is it still relevant?

HannesSommer:
#70 we have several options:

1. This PR
2. Fix  #66 and then roll back to using just yaml-cpp-pm as library name
3. Pack yaml into libpiontmatcher (not treating it as dep anymore)
4.  Migrate to yaml API 0.5

HannesSommer:
@pomerlef , maybe you should merge this in till there is time for a better solution. It should not do any harm (I'm pretty positive) and solve the issue #70 for all users that use libpointmatcher as dynamic library - including the ros build farm, which would effectively render the CI done by the ros farm useful again helping to detect new regressions and so on .. :).

PR #79, opened 2015-04-11
protected adding EXTRA_DEPS as dependencies against the variable being empty (further addressing #70).

HannesSommer:


pomerlef:
thanks!

Issue #42, opened 2014-07-24
Update to newer yaml-cpp API


samlcharreyron:
Yaml-cpp has a new API since version 0.5.  Newer versions of Ubuntu now use yaml-cpp0.5 by default so libpointmatcher should be updated to use the new API.

lingerer:
I try to use system yaml api and found the same problem.I check the code and I thought thie issue should solve.

AlexandreG87:
I am currently working on updating yaml-cpp to its new API. So far, I've managed to modify libpointmatcher's code to make it compile and work with it. There were only a few changes to be made. For now, it works with the following configurations:

| yaml-cpp version | Ubuntu version |
| :---: | :---: |
| 0.5.2 | Ubuntu 18.04.5 |
| 0.6.2 | Ubuntu 20.04.1 |

As mentioned in the first post, the new yaml-cpp API is available since at least Ubuntu 16.04, i.e. version 0.5.2.

So, for Ubuntu users, it seems to me that it would be simpler to just use the system installed yaml-cpp.

Same thing for macOs users, it is possible to install version 0.6.3 with homebrew.  So I don't think that would be a problem.

On the other hand, Windows users will have to install yaml-cpp from the source files. This is probably better for troubleshooting and debugging purposes.

Before creating a pull-request, I am asking for your thoughts about my proposal, which is to no longer provide the source code of yaml-cpp and let libpointmatcher users install or compile yaml-cpp version 0.5.2 and higher by themselves.

Thanks!

pomerlef:
That's a though one. Adding the sources internally cuts us from external update so
- we loose bug updates, but
- the library doesn't break everytime the external dependence is updated
- we will have to deal with a spread of versions depending on the OS

@simonpierredeschenes and @YoshuaNava any thoughts from a CS perspective?

aguenette:
>     * we loose bug updates, but
>     * the library doesn't break everytime the external dependence is updated
>     * we will have to deal with a spread of versions depending on the OS

If it can help make a decision, I would like to add that the new yaml-cpp API has been around for just over eight years now. So, knowing that it has not changed between versions 0.5.0 and 0.6.3, I think it might be possible to leave it up to the user to install a compatible version of yaml-cpp. What do you think?

pomerlef:
alright, let's got with that then!

PR #87, opened 2015-05-06
Fix/invasive yaml namespace redirection

HannesSommer:
With this PR libpointmatcher will not only support linking to a different YAML library but also have one included together with it in a CU. 
It needed to not expose the namespace aliasing (the alias YAML is now below PointMatcherSupport) and changing the include guards.
Of course all this yaml contrib is a nightmare but till #42 is closed this is probably the best we can do. 
Or are there any better fundamental ideas?


rdube:
+1 Thanks for working on this Hannes!

pomerlef:
I'm in the middle of something. Unless it's urgent, I'll check that next week.

@HannesSommer Thanks indeed for looking into that.

HannesSommer:
No problem. It has time.

* 3884cb4 is large but trivial (automatic renaming of include guards) and will do no harm.
* I'm pretty positive that 9238a6a is a pure improvement.
So if time is precious, I would suggest merging without any further investment.

PR #90, opened 2015-05-17
Small fix to docs

wolfv:
registrar.cpp --> Registry.cpp


pomerlef:
Thanks for taking the time to do a pull request instead of just reporting the error. It's appreciated.

Issue #101, opened 2015-07-05
Missing DESTINATION for TARGETS error when building libpointmatcher for Windows


kwill:
I have been able to build all of the dependencies successfully on Windows. Now that I am trying to build libpointmatcher itself, I am getting the following error:

````
-- libnabo found, version 1.0.6
-- using built-in yaml-cpp, version 0.3.0
   -- text-based configuration enabled
CMake Error at CMakeLists.txt:262 (install):
  install Library TARGETS given no DESTINATION!
````

Please let me know what other information I can provide.

**Comment by pomerlef** [comment]
Could you send the output of that command (from your build directory):
```
cmake -LAH
```
It's either the variable `CMAKE_INSTALL_PREFIX` or `INSTALL_LIB_DIR` that is empty. Given that the goal is to produce a `.sln` file, it shouldn't matter.

It might worth to consider having a `.sln` directly on Github. How other projects release their sources on Windows?

**Comment by kwill** [comment]
Thanks for the feedback, LAH output follows and I can already identify several issues. I'm not sure what the typical or best approach is, but I have had some issues with prebaked `.sln` files in the past, so building them myself seems safer. Also, your instructions are very clear, which makes it easier.

```
CMake Error at CMakeLists.txt:25 (MESSAGE):
  32 bits compiler detected.  Libpointmatcher is only supported in 64 bits.


-- Could NOT find Doxygen (missing:  DOXYGEN_EXECUTABLE) 
CMake Error at C:/Program Files (x86)/CMake/share/cmake-3.2/Modules/FindBoost.cmake:1182 (message):
  Unable to find the requested Boost libraries.

  Unable to find the Boost header files.  Please set BOOST_ROOT to the root
  directory containing Boost or BOOST_INCLUDEDIR to the directory containing
  Boost's headers.
Call Stack (most recent call first):
  CMakeLists.txt:106 (find_package)


-- libnabo found, version 1.0.6
CMake Warning (dev) at CMakeLists.txt:191 (get_property):
  Policy CMP0026 is not set: Disallow use of the LOCATION target property.
  Run "cmake --help-policy CMP0026" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.

  The LOCATION property should not be read from target "yaml-cpp-pm".  Use
  the target name directly with add_custom_command, or use the generator
  expression $<TARGET_FILE>, as appropriate.

This warning is for project developers.  Use -Wno-dev to suppress it.

[...]

PR #102, opened 2015-07-14
Make Windows compilation instructions a bit more explicit

kwill:
These are some minor tweaks to the Windows compilation instructions for anyone (like me) less familiar with the toolchain.

Note that I am currently having an SLN generation issue (https://github.com/ethz-asl/libpointmatcher/issues/101) with libpointmatcher itself, so the instructions for the final step should be double-checked for correctness.

### DIFF
--- a/doc/CompilationWindows.md
+++ b/doc/CompilationWindows.md
@@ -34,7 +34,7 @@ grep --version
 ```
 
 ### Building Boost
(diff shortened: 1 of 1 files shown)


kwill:
Here are some images if you want to include them in the documentation.

cmake_yamlcpp

cmake_libnabo

cmake_libpointmatcher

pomerlef:
Thanks for the update, it's very appreciated. I'll add your screenshots.

pomerlef:
The images are in and I added you in our Special Thanks section.

PR #103, opened 2015-07-28
Added RandomSampling filter from PCL. Fixed loading ply files with color information. Improved Histogram stats.

carlosmccosta:
Tested in Ubuntu 14.04 with ROS Indigo for comparison with DRL system:
https://github.com/carlosmccosta/dynamic_robot_localization_tests/blob/hydro-devel/launch/environments/asl/bags/ethzasl_kinect_dataset_high_complexity_slow_fly_movement.launch

roslaunch dynamic_robot_localization_tests ethzasl_kinect_dataset_high_complexity_slow_fly_movement.launch use_dynamic_robot_localization:=0 use_ethzasl_icp_mapper:=1

To integrate with ROS Indigo ethzasl_icp_mapper i had to compile with:
git clone git@github.com:carlosmccosta/libpointmatcher.git
mkdir libpointmatcher_build && cd libpointmatcher_build
cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/opt/ros/indigo ../libpointmatcher
make
sudo make install

### DIFF
--- a/pointmatcher/DataPointsFiltersImpl.cpp
+++ b/pointmatcher/DataPointsFiltersImpl.cpp
@@ -1059,6 +1059,11 @@ DataPointsFiltersImpl<T>::MaxPointCountDataPointsFilter::MaxPointCountDataPoints
 	RandomSamplingDataPointsFilter("MaxPointCountDataPointsFilter", MaxPointCountDataPointsFilter::availableParameters(), params),
 	maxCount(Parametrizable::get<unsigned>("maxCount"))
 {
+	try {
+		seed = Parametrizable::get<unsigned>("seed");
+	} catch (const InvalidParameter& e) {
+		seed = static_cast<unsigned int> (time (NULL));
+	}
 }
 
 // Compute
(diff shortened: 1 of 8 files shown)


pointmatcher/DataPointsFiltersImpl.h
```diff
@@ -405,6 +407,7 @@ struct DataPointsFiltersImpl
 		}
 		
 		const unsigned maxCount;
+		unsigned seed;
```
pomerlef: Don't forget to add this parameter in the ParametersDoc structure (line 405).

pointmatcher/DataPointsFiltersImpl.h
```diff
@@ -37,6 +37,8 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #define __POINTMATCHER_DATAPOINTSFILTERS_H
 
 #include "PointMatcher.h"
+#include "time.h"
+#include "stdlib.h"
```
pomerlef: I think those two includes (time.h and stdlib.h) will cause problem when compile on Windows. If equivalent function are in Boost, it's better. We had to revert back a lot of those to maintain compatibility a while ago. I would prefer to not have to do that again.

pointmatcher/DataPointsFiltersImpl.cpp
```diff
@@ -1059,6 +1059,11 @@ DataPointsFiltersImpl<T>::MaxPointCountDataPointsFilter::MaxPointCountDataPoints
 	RandomSamplingDataPointsFilter("MaxPointCountDataPointsFilter", MaxPointCountDataPointsFilter::availableParameters(), params),
```
pomerlef: this doesn't seem to be needed anymore given your new implementation.

pointmatcher/IO.cpp
```diff
@@ -1524,6 +1524,10 @@ typename PointMatcherIO<T>::DataPoints PointMatcherIO<T>::loadPLY(std::istream&
 			const int row = vertex->properties[propID].pmRowID;
 			const PMPropTypes type = vertex->properties[propID].pmType;
 			
+			if (vertex->properties[propID].name == "red" || vertex->properties[propID].name == "green" || vertex->properties[propID].name == "blue" || vertex->properties[propID].name == "alpha") {
```
pomerlef: This seems to make one of the unit tests fail. If it's doing what it should, modify the unit test accordingly.

pomerlef:
Thanks for the patch, those seems valuable add-ons. 

I won't merge it as it is as the patch will break other systems. I commented part of your commits where I have concerns. In general, break your merge requests in smaller, thematic chunk so it's easier to review and quicker to approve.

Thanks!

pointmatcher/DataPointsFiltersImpl.h
carlosmccosta: Added new "seed" parameter and removed previous "prob" parameter.

pointmatcher/IO.cpp
carlosmccosta: The unit test was failing because the savePLY had the same problem as the loadPLY.
Analyzing the IO class, it seems that colors are assumed to be in range [0,1] in float format.
But ply files store each channel as an unsigned char in range [0,255].
The modifications i introduced take this ply format detail in consideration.

carlosmccosta:
Thanks for the feedback :)
Last commit addressed the issues related with the unit tests for ply IO and the time and stdlib headers.
Have a nice day :)

pomerlef:
Great, everything compiled and passed on Mac too. 

Thanks for the update.

PR #110, opened 2015-08-21
Removed dead cmake code

HannesSommer:
I don't think these lines are doing anything. 
Other opinions?


pomerlef:
That was part of Cmake original tutorial on how to make a ProjectConfig:
http://www.cmake.org/Wiki/CMake/Tutorials/How_to_create_a_ProjectConfig.cmake_file

It seems to be removed from their updated tutorial.

PR #115, opened 2015-09-16
Handle the degenerate situation when the residuals are 0

oleg-alexandrov:
I found that point-to-plane alignment fails when one tries to align a point cloud to itself. The reason is that all the residuals are zero, and Eigen::AxisAngle when applied to a zero vector returns an invalid rotation, full of NaNs. The fix is to just return the identity alignment transform.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ The library has a light dependency list:
 
  * [Eigen] version 3, a modern C++ matrix and linear-algebra library,
  * [boost] version 1.48 and up, portable C++ source libraries,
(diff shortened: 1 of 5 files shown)


oleg-alexandrov:
Actually, I need to also test the case when there is pure translation and no rotation (rather than just zero translation and no rotation). Some unit tests should be added as well. I will look at this tomorrow. So for now this pull request is not ready.

pomerlef:
ok no problem, I'll wait. Thanks for digging into that one!

PR #117, opened 2015-09-17
Handle the degenerate situation when the residuals are 0.

oleg-alexandrov:
I wiped the previous pull request, and made a new one to be in sync with the latest upstream changes (I could have just appended to that request I guess, but it is cleaner this way).

To summarize, the bug here actually shows up very rarely. I did add a unit test which reproduces it. In order for this issue to show up, one must use point-to-plane alignment, have the two clouds to align be identical, and any filtering which removes points must be turned off. Then, the Eigen::AngleAxis() function cannot create a rotation from a zero vector. The correct rotation is the identity though.

The bug does not show up if one cloud is translated in respect to the other one. Numerical noise results in a non-zero (but very small) rotation matrix.

The bug does not show up either with point-to-point alignment when the clouds are identical.

I fixed one more small issue. The function Matches::getDistsQuantile(const T quantile) did not return the correct answer if quartile = 1.0. That was not issue until now, but could be from now on, since the ratio in filtering is now allowed to be 1 per the previous fix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ The library has a light dependency list:
 
  * [Eigen] version 3, a modern C++ matrix and linear-algebra library,
  * [boost] version 1.48 and up, portable C++ source libraries,
- * [libnabo] version 1.0.1, a fast K Nearest Neighbour library for low-dimensional spaces,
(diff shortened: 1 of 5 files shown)


pomerlef:
Thanks for the contributions, it's very appreciated.

PR #120, opened 2015-09-23
fix compilation of utest, missing pthread library when linking

bkueng:
Hi
This fixes come unresolved symbols to pthread when linking under Fedora 22.


HannesSommer:
Thanks! Looks good to me.

PR #121, opened 2015-09-23
fix PointMatcher: SVD can return a reflection instead of a rotation

bkueng:
Hi
The SVD in the point matcher can return a reflection instead of a rotation (det=-1 instead of det=1). This later causes the exception "Error, rotation matrix is not orthogonal." due to det(rotationmatrix) = -1.
This fixes it by taking the second best solution of the SVD, guaranteed to be a rotation.
see also: http://igl.ethz.ch/projects/ARAP/svd_rot.pdf


pomerlef:
Great path. thanks!

PR #127, opened 2015-11-02
Fixed warnings previously introduced

gawela:
Removed some unused expressions.


pomerlef:
Awesome, thanks a lot!

PR #132, opened 2016-01-12
Support binary VTK legacy files

HannesSommer:
NOT READY to MERGE. 

I just needed faster writing of point clouds and this quick and dirty implementation (not polished for merging yet) is already about 20 times faster than the ASCII version. And e.g. paraview is reading it way faster.

@pomerlef , do you think this is a useful extension (even though it is legacy VTK files anyway and reading the same files would not be supported without further effort) or would you rather switch to libvtk to read and write vtk files. Was there a good reason not to do that in the first place?

### DIFF
--- a/pointmatcher/IO.cpp
+++ b/pointmatcher/IO.cpp
@@ -887,14 +887,18 @@ void PointMatcher<T>::DataPoints::save(const std::string& fileName, bool binary)
 	const string& ext(boost::filesystem::extension(path));
 	if (boost::iequals(ext, ".vtk"))
 		return PointMatcherIO<T>::saveVTK(*this, fileName, binary);
-	else if (boost::iequals(ext, ".csv"))
+
+	if (binary)
+		throw runtime_error("save(): Binary writing is not supported together with extension \"" + ext + "\". Currently binary writing is only supported with \".vtk\".");
+
+	if (boost::iequals(ext, ".csv"))
 		return PointMatcherIO<T>::saveCSV(*this, fileName);
 	else if (boost::iequals(ext, ".ply"))
 		return PointMatcherIO<T>::savePLY(*this, fileName);
 	else if (boost::iequals(ext, ".pcd"))
 		return PointMatcherIO<T>::savePCD(*this, fileName);
 	else
-		throw runtime_error("saveAnyFormat(): Unknown extension \"" + ext + "\" for file \"" + fileName + "\", extension must be either \".vtk\" or \".csv\"");
+		throw runtime_error("save(): Unknown extension \"" + ext + "\" for file \"" + fileName + "\", extension must be either \".vtk\", \".ply\", \".pcd\" or \".csv\"");
 }
 
 template
--- a/pointmatcher/IO.cpp
+++ b/pointmatcher/IO.cpp
@@ -986,6 +986,28 @@ typename PointMatcher<T>::DataPoints PointMatcherIO<T>::loadVTK(const std::strin
 	return loadVTK(ifs);
 }
 
+void skipBlock(bool binary, int binarySize, std::istream & is, bool hasSeparateSizeParameter = true){
+	int n;
+	int size;
+	is >> n;
+	if(hasSeparateSizeParameter) {
+		is >> size;
+	} else {
+		size = n;
+	}
+
+	std::string line;
+	getline(is, line); // remove line end after parameters;
+	if(binary){
+		is.seekg(size * binarySize, std::ios_base::cur);
+	} else {
+		for (int p = 0; p < n; p++)
+		{
+			getline(is, line);
+		}
+	}
+}
+
 //! Load point cloud from a stream as VTK
 template<typename T>
 typename PointMatcher<T>::DataPoints PointMatcherIO<T>::loadVTK(std::istream& is)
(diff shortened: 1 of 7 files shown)


pomerlef:
@HannesSommer : the general strategy is to keep the dependency list as small as possible. We provide basic functionalities from inside the library and, for more advance features, we create a binder like:
- [pointmatcher-ros](https://github.com/ethz-asl/pointmatcher-ros): pull out by Ralf
- [libpointmatcher_vtk](https://github.com/ethz-asl/libpointmatcher_vtk): I've done that for Alstom, there is not much in it.

That's the general guideline but, as you can see, we didn't went too far in that direction and it's apparently not standardize for the naming. That being said, everything that increase the speed of the workflow is good. If you are in the order of 20x then we should take the time to polish that. This feature was requested a couple of time in the issue list too.

Is it a lot of work to support it both way (read/write)? With the trend of issue I saw, it will confuse a lot of users if not.

HannesSommer:
Ah, thanks for the quick reply. I totally get the aim of small dependency footprint. It is actually probably an 1h of work for me to polish it enough and add write support - at least for most of the data (see below). I can do that.

A few questions remain, though:

1. Colour: 
How important is storing colour? (see https://github.com/ethz-asl/libpointmatcher/pull/132/files#diff-65f29f99818f583e6f26c71cfaade3cdR505). Would it be fine to start with binary only for point clouds without colour feature? That would mean falling back to ASCII if there is colour in there. Adding colour would be 30min extra I suppose and could in theory be done later.

2. new save binary flag:
What about this interface extension https://github.com/ethz-asl/libpointmatcher/pull/132/files#diff-5cc32d1d2c455d67872c78be20145a26R201. Is that fine, would you do it different? This gets passed down till here https://github.com/ethz-asl/libpointmatcher/pull/132/files#diff-03f31f751a0bac300a04bd031c66ec48R878. The fact that the others don't support binary I don't know how to treat the libpointmatcher way :). What would it be? Throw an exception (like binary not supported for the BLA format) or silently or with a warning switch to non binary if binary is not supported.

3. New writeBinary parameter:
Same question here:https://github.com/ethz-asl/libpointmatcher/pull/132/files#diff-d702c135dcab9ea187a2353e809ffd82R168

4. ASCII data type (float or double):
In this line https://github.com/ethz-asl/libpointmatcher/pull/132/files#diff-65f29f99818f583e6f26c71cfaade3cdR210 there is currently a change of the existing semantic for the ASCII output: in case of ```T``` equals ```double``` it would write ```double``` instead of ```float```. Nothing else would change for ASCII output. However, if such a file gets read e.g. with paraview it would probably read a double point cloud instead of a float pointcloud afterwards. Why did you hard code ```float``` there? (I think currently libpointmatcher will just use T when reading ASCII data - no matter whether there is float or double written in the vtk file)

pomerlef:
1. ok
2. throw an exception with the description of the error and what to do.
3. see 2
4. just like that, I don't recall why `float` is hard coded. It seems better to use `T` so try it like that. Test if there is a problem in Paraview for `double` and `float` formats, just to make sure.

Thanks for looking into that. The IO.cpp is halfway through a refactoring process (since 1-2 years), so it's a bit of a mess. At some point, I will be bored enough to finish that...

pomerlef:
It's the second time I'm seeing a unit test failure on `DataFilterTest.ElipsoidsDataPointsFilter`. It works on my MacOS machine so I didn't catch it before. It seems to fail on my Ubuntu machine.

@HannesSommer do you observe the same thing?
@rdube was that test working when you added it? Is it a recent thing?

HannesSommer:
I think the result is random. Try 
```./utest/utest --gtest_filter=DataFilterTest.ElipsoidsDataPointsFilter --gtest_repeat=100 --path <POINTMATCHER_SRC>/example/data/``` 
and it will fail sometimes.

HannesSommer:
Probably some thresholds need adapted or the random corner cases need to get filtered out.

pomerlef:
ouuu lucky me. Indeed, it fails in two ways:
1. ```unknown file: Failure
C++ exception with description "no outlier to filter" thrown in the test body.```
2. ```The difference between validTrans and testTrans is 0.059576287865638733, which exceeds 0.05, where
validTrans evaluates to 0.22068966925144196,
testTrans evaluates to 0.28026595711708069, and
0.05 evaluates to 0.050000000000000003.```

HannesSommer:
Finally I've found some time to but some work in this.
I've tried to implement all your suggestions.

I've reached the feature set of the VTK ASCII part (so reading writing including colour). 
I've amended the unit test with a VTKBinary test. 

The correctness of the written files I've tried to validate a bit with paraview. But this wasn't near to heavy testing! In my normal work flow I don't use colours and I don't read VTK with pointmatcher at all (only paraview). So the question is how do we test this? Can we mark this as a experimental feature? Do you, @pomerlef , have normal work that involves colour and reading VTK?

pomerlef:
Me personally, no. It's mainly for people using the Kinect. I'll send you a file by mail in ASCII with a color field to test. The upload attachment size is limited to 10MB here.

pomerlef:
If it's only few corner cases with the colors that you're not sure about, I wouldn't mind to add it in the main branch and deal with deal with bugs that might raise from it later. If somebody finds a bug, that means that he was in a better situation than us to test it.

PR #134, opened 2016-01-14
syedharoonalam patch

syedharoonalam:


pomerlef:
Thanks. I removed the typedef in icp_advance_api. This was redundant given that the code was starting with:
```
 using namespace PointMatcherSupport;
```
Interesting how different compilers catch different things.

PR #135, opened 2016-01-19
Weighted error minimizer

bkueng:
Hi François
This extends the error minimizer to use weighted least squares minimization. It requires an appropriate OutlierFilter that outputs weights in [0, 1]. With {0, 1} weights, the behavior is unchanged.

- runtime performance similar as before
- unit test passing
- based on: http://igl.ethz.ch/projects/ARAP/svd_rot.pdf (not umeyama 91, this is for the scaling)


pomerlef:
Great contribution. Thanks for looking into that and cleaning the old code.

Issue #67, opened 2015-02-28
RemoveNaNDataPointsFilter crashes on Mac OS X...


drhalftone:
I finally have libpointmatcher up and running and am able to do absolutely awesome alignments if I manually remove NANs, but when I allow NANs in my point clouds and enable the RemoveNaN filter, the software consistently crashes. Here is the exception message from the application log:

libc++abi.dylib: terminating with uncaught exception of type PointMatcher<float>::ConvergenceError: no outlier to filter

drhalftone:
I also found a second situation that creates a similar error.  This time, I tried to align a point cloud with itself.  So my guess is that the distances between points is going to be zero for all points, and the outliers filter (outlierFilters:  - TrimmedDistOutlierFilter: ratio: 0.2) can't decide which points to keep and which to delete.  Seems like you would have to make a random decision to keep points and remove others in cases of a tie.

pomerlef:
could you send one test file so we could replicate on our side?

drhalftone:
I’ll send you one as soon as I get the chance (in a day or two).


On Apr 20, 2015, at 4:48 PM, François Pomerleau <notifications@github.com> wrote:

could you send one test file so we could replicate on our side?

pomerlef:
If there is no follow-up on this, I'll close it.

PR #138, opened 2016-01-28
Calculating mean using filtered data to avoid NaNs

vanurag:
In the presence of NaNs in reference data, program crashes even after using RemoveNaNDataPointsFilter since the mean is being calculated on non-filtered data. Also addresses the issue #67.


pomerlef:
Great, looks good. Thanks for looking into that!

PR #145, opened 2016-03-12
Fix for a no-longer valid yaml file

oleg-alexandrov:
I found a yaml file which no longer satisfies the existing syntax. I put a fix. I set knn to 10, as that's what is mostly used in other yaml files.


pomerlef:
Thanks for the update. I'll put the knn value back to 7 as this file was used to generated benchmark results.

PR #146, opened 2016-03-12
Add yaml file testing point-to-point icp

oleg-alexandrov:
There are a lot of yaml files that are used in unit tests, but none test point-to-point icp, they are all for point-to-plane. I added one, to make the test sequence more complete.


oleg-alexandrov:
I have no idea why the test failed. It worked for me. This is a trivial commit and whatever problem happened, should be easy to fix. I don't have access to the log above to see the problem.

HannesSommer:
retest this please

HannesSommer:
The error message was 
```
/var/lib/jenkins/jobs/libpointmatcher/workspace/utest/utest.cpp:147
Expected: (transErr) < (transTol), actual: 0.144691 vs 0.1
   ../examples/data/icp_data/defaultPointToPointMinDistDataPointsFilter.yaml
```
I let it rerun. But the result is the same. It does not seem to be random and unlikely. It doesn't fail on your computer?

oleg-alexandrov:
I related a bit the translation check. Now it passes.

pomerlef:
thanks Oleg

PR #147, opened 2016-03-16
Similarity

oleg-alexandrov:
Integrated the similarity transform code. Added two unit tests. Want to see if the tests pass on the server.


oleg-alexandrov:
The tests failed again, though they pass for me on two separate machines. I wish I could see the error message.

ffurrer:
The error is:

```
/var/lib/jenkins/jobs/libpointmatcher/workspace/utest/utest.cpp:151: Failure
Expected: (transErr) < (transTol), actual: 0.166578 vs 0.15
   ../examples/data/icp_data/defaultPointToPointMinDistDataPointsFilter.yaml
[  FAILED  ] icpTest.icpTest (6051 ms)
```

oleg-alexandrov:
I revamped completely how a test is judged to pass. The tests compare the reference ICP transform saved on disk vs the new ICP transform computed with the current code. The problem is that ICP does not have a unique solution, and on different machines different answers are obtained. 

Yet, the point of icp is curT*data = ref, and refT*data = ref (here curT is the recent transform and refT is the old version stored on disk). 

So no matter what, the difference curT*data - refT*data must be small. The new metric I test for is then 

median(abs(curT*data - refT*data)) / median(abs(curT*data)) < 0.03.

This seems to work well on two Linux systems I tested, on one Mac system, and your own server.

pomerlef:
Thanks for the addon!

pointmatcher/ErrorMinimizersImpl.h
```diff
@@ -76,11 +76,22 @@ struct ErrorMinimizersImpl
 		virtual T getOverlap() const;
 	};
 
+  	struct PointToPointSimilarityErrorMinimizer: ErrorMinimizer
+	{
+		inline static const std::string description()
+		{
+			return "Point-to-point similarity error (rotation + translation + scale). Based on SVD decomposition. Based on \\cite{Umeyama1991}.";
```
pomerlef: If I understood well, the scale computed here is equal for all dimensions. The term similarity is appropriate here, but extra explanation would avoid users confusing it with affine.

PR #150, opened 2016-03-17
Minor fixes

oleg-alexandrov:
I clarified the similarity transform, per Francois. And I found what I think is a bug. A typo prevented an input variable from being passed to a class, instead, the class member was assigned to itself. The compiler was complaining.

### DIFF
--- a/pointmatcher/ErrorMinimizersImpl.h
+++ b/pointmatcher/ErrorMinimizersImpl.h
@@ -69,7 +69,7 @@ struct ErrorMinimizersImpl
 	{
 		inline static const std::string description()
 		{
-			return "Point-to-point error. Based on SVD decomposition. Based on \\cite{Besl1992Point2Point}.";
+			return "Point-to-point error. Based on SVD decomposition. Per \\cite{Besl1992Point2Point}.";
 		}
 		
 		virtual TransformationParameters compute(const DataPoints& filteredReading, const DataPoints& filteredReference, const OutlierWeights& outlierWeights, const Matches& matches);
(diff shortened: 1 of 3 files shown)


oleg-alexandrov:
From the diff it is not clear what the bug is about. It happens in the line that says:
gravity(gravity)
which did not change. Yet the input variable was gravitity, and was not used. Now it is renamed to gravity and is used.

I find it quite confusing that the input variable names are the same as class member names. That is not a nice practice.

pomerlef:
Strange, that part of the code is very old, but there was clearly a typo there. The message is not clear as whether the patch fix or not the problem. Do I need to investigate this?

As for the convention for using the same input and class member, I agree that it can be confusing. In theory, we use that mainly when the constructor just copies the inputs to the member as they are the same concept and discarded right away. I know that some conventions use a suffix to make to tag members like `m_variableName`, but I find that heavy. If there is a doubt in the function call, I tend to use `this->variableName`. I'm not sure if some conventions tag input variable instead.

oleg-alexandrov:
Yes, my patch fixes the problem. You can just read the whole block of code, up to gravity(gravity) before my fix, and after the fix. Again, the input is has a typo: gravitity instead of gravity. 

If I were to write that code I would use gravity_in as input, and gravity as member variable, so the invocation would be gravity(gravity_in). Or even the m_gravity notation that you are not happy with.

pomerlef:
ok great, thanks for the patch.

I'll open an issue about the unclarity of the inputs name and tag it as enhancement. It might need to wait until I've access to another intern as changing convention is a long reformating task (but we might discover other bugs by doing so).

PR #154, opened 2016-04-04
Add getResidualError methods to ErrorMinimizer

Ellon:
I added the getResidualError method to be able to assess the ICP result through the residual error. I'm sharing it here in case you're interested to add the functionality to the library.

Cheers!

----- Begin Commit Message -----
[error minimizer] Add getResidualError methods

getResidualError is a const method and makes use of getMatchedPoints, thus we
also add a const version of getMatchedPoints that does not change
ErrorMinimizer's lastErrorElements.

The residual error is computed in a separated function defined in an anonymous
namespace


pomerlef:
Don't despair, I hope to be able to include your patch within this month...

pomerlef:
Alright, I had to do quite some changes to avoid code duplication, but your computation for error residuals are now in.

Thanks!

Issue #162, opened 2016-06-24
Throw exception when reaching max number of iterations


Ellon:
I believe that when computing the transformation with ICP an exception should be thrown then reaching the max number of allowed iterations. 

AFAIK currently it is not possible to know if the ICP computation stopped because we reached the max number of iterations or because other checker signalized that the ICP converged. Throwing an exception that contains with the last computed transformation would give the option to the user to discard the result (since in most of the cases this will mean that the ICP didn't converge) or use it anyway, if the computed transformation is still relevant for the application.

pomerlef:
I'm adding the same comment I did to your pull request here, as I think it make more sense to have this discussion here.

There is a small link with the issue #23 here.

There seem to be a general need to have access to more information through the API about the iterative process either _live_ or after the iterations. So, I agree with @HannesSommer that there is some planning to be done here.

It would make sense to collect here what kind of information users want to access and list so we can rethink properly the `TransformationChecker` and `Inspector` class.

Issue #23, opened 2013-12-20
Fine-grained control of inspector module


oleg-alexandrov:
ICP being an iterative algorithm, I think it can be very important for the user (and developer) to see how it performs on given data set. For example, to fine-tune the stopping criteria, the user may want to know the iteration info, that is, how the error is decreasing (does the error decrease only for 10 iterations, and for the next 90 just some wiggles happen, so 10 iterations are enough?).

libpointmatcher provides for that the Inspector module. However, this module dumps everything, including all point clouds at every single iteration, which can be huge. That is good for debugging, but the user may want to control what info to dump, for performance reasons. 

I implemented this fine-level control of Inspector in my own copy of libpointmatcher, in a way that should be easy to merge should you desire that. I needed this for my alignment tool which uses libpointmatcher.

And by the way, I fixed a bug with the name of the columns in the iteration info file, the translation and rotation column names were reversed, here's the fix: https://github.com/oleg-alexandrov/libpointmatcher/commit/54c9da84c1f7cc1098f5d1d9ecb4427bbe5d3563

And here's the entire set of fixes: https://github.com/ethz-asl/libpointmatcher/pull/22

PR #163, opened 2016-06-24
[feature] Throws if max allowed iteractions is reached.

Ellon:
Implements a solution for #162

### DIFF
--- a/pointmatcher/Exceptions.cpp
+++ b/pointmatcher/Exceptions.cpp
@@ -41,7 +41,11 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 template<typename T>
 PointMatcher<T>::ConvergenceError::ConvergenceError(const std::string& reason):
 	runtime_error(reason)
-{} 
+{}
+ 
+template<typename T>
+PointMatcher<T>::ConvergenceError::~ConvergenceError() throw ()
+{}
 
 template struct PointMatcher<float>::ConvergenceError;
 template struct PointMatcher<double>::ConvergenceError;
--- a/pointmatcher/TransformationCheckersImpl.h
+++ b/pointmatcher/TransformationCheckersImpl.h
@@ -57,6 +57,12 @@ struct TransformationCheckersImpl
 	
 	struct CounterTransformationChecker: public TransformationChecker
 	{
+		//! ICP reached max number of iterations
+		struct MaxNumIterationsReached: std::runtime_error
+		{
+			MaxNumIterationsReached();
+		};
+
 		inline static const std::string description()
 		{
 			return "This checker stops the ICP loop after a certain number of iterations.";
--- a/pointmatcher/ICP.cpp
+++ b/pointmatcher/ICP.cpp
@@ -60,6 +60,19 @@ InvalidModuleType::InvalidModuleType(const std::string& reason):
 	runtime_error(reason)
 {}
 
+template<typename T>
+PointMatcher<T>::ICPChainBase::MaxNumIterationsError::MaxNumIterationsError(const std::string& reason, const TransformationParameters& Tlast):
+ConvergenceError(reason),
+lastTransformation(Tlast)
+{}
+
+template<typename T>
+PointMatcher<T>::ICPChainBase::MaxNumIterationsError::~MaxNumIterationsError() throw ()
+{}
+
+template struct PointMatcher<float>::ICPChainBase::MaxNumIterationsError;
+template struct PointMatcher<double>::ICPChainBase::MaxNumIterationsError;
+
 //! Protected contstructor, to prevent the creation of this object
 template<typename T>
 PointMatcher<T>::ICPChainBase::ICPChainBase():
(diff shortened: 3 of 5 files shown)


Ellon:
Er... I can't see the details of why it failed. :frowning:

Ellon:
Ok, no problem for waiting.

I don't think I have false positives: what I did was just to throw a "[internal exception](https://github.com/ethz-asl/libpointmatcher/pull/163/files#diff-7528c4724ece65d996b348f0989c31ebR82)" at the same point `CounterTransformationChecker` would set `iterate = false`, then I [catch it](https://github.com/ethz-asl/libpointmatcher/pull/163/files#diff-fe51353a762bb8985688217a15b0d0c9R419) inside `ICP::computeWithTransformedReference`, and return the same computed transformation through an [exception](https://github.com/ethz-asl/libpointmatcher/pull/163/files#diff-fe51353a762bb8985688217a15b0d0c9R447), instead of through the method's return value. 

Maybe some of the tests were conceived without considering that the ICP would reach the max iterations when running them? 

Also, here I have only 4 failures, instead of the 6 failures from your jenkins build: 
```
$ utest/utest --path ../examples/data/
(...)
[----------] Global test environment tear-down
[==========] 58 tests from 13 test cases ran. (28604 ms total)
[  PASSED  ] 54 tests.
[  FAILED  ] 4 tests, listed below:
[  FAILED  ] DataFilterTest.MaxDensityDataPointsFilter
[  FAILED  ] MatcherTest.KDTreeMatcher
[  FAILED  ] TransformationCheckerTest.CounterTransformationChecker
[  FAILED  ] TransformationCheckerTest.BoundTransformationChecker
```
And, if I run the `DataFilterTest.MaxDensityDataPointsFilter` test separately I don't have a failure:
```
$ utest/utest --path ../examples/data/ --gtest_filter=DataFilterTest.MaxDensityDataPointsFilter
Note: Google Test filter = DataFilterTest.MaxDensityDataPointsFilter
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from DataFilterTest
[ RUN      ] DataFilterTest.MaxDensityDataPointsFilter
[       OK ] DataFilterTest.MaxDensityDataPointsFilter (1001 ms)
[----------] 1 test from DataFilterTest (1001 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (1001 ms total)
[  PASSED  ] 1 test.
```
Maybe something is wrong with multiple tests being performed in sequence?

pointmatcher/Exceptions.cpp
```diff
@@ -41,7 +41,11 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 template<typename T>
 PointMatcher<T>::ConvergenceError::ConvergenceError(const std::string& reason):
 	runtime_error(reason)
-{} 
+{}
+ 
+template<typename T>
+PointMatcher<T>::ConvergenceError::~ConvergenceError() throw ()
```
HannesSommer: This is not required and not related to the PR. I would drop it. Of course also the corresponding declaration below. (The default destructor is already virtual and equivalent to this one because ConvergenceError has a base class with a virtual destructor.)

pointmatcher/TransformationCheckersImpl.h
```diff
@@ -57,6 +57,12 @@ struct TransformationCheckersImpl
 	
 	struct CounterTransformationChecker: public TransformationChecker
 	{
+		//! ICP reached max number of iterations
+		struct MaxNumIterationsReached: std::runtime_error
```
HannesSommer: This class neither needs to be child std::runtime_error IMO.

pointmatcher/TransformationCheckersImpl.h
```diff
@@ -57,6 +57,12 @@ struct TransformationCheckersImpl
...
+		{
+			MaxNumIterationsReached();
```
HannesSommer: Not required.

HannesSommer:
I though a bit about your suggestion. I totally appreciate the idea of notifying the user somehow of an incomplete convergence. This should be supported.

But having a special exception kind of preserving the otherwise returned transformation seems to me the wrong way. Furthermore this would be inconsistent with how `BoundTransformationChecker` does it. See https://github.com/ethz-asl/libpointmatcher/blob/master/pointmatcher/TransformationCheckersImpl.cpp#L214.

Mostly because this new exception can actually break running systems I would not merge in an unconditional throw at max iterations.

Instead for both transformation checkers I would suggest to not throw or at least to introduce a boolean flag, with which the user can decide whether / which of these types of end of iteration should trigger an exception. And if not the premature transformation gets returned (for those users who still want to use the result). For those users to still be able to distinguish the different kinds of "conversion" something else should be in place. Some data structure where the reasons to stop ICP can be collected and queried. 

But to introduce this doesn't seem like a minor fix and requires some planning especially because of the very liberal concept of a list of transformation checkers. 

Other opinions?

pomerlef:
There is a small link with the issue #23 here.

There seem to be a general need to have access to more information through the API about the iterative process either _live_ or after the iterations. So, I agree with @HannesSommer that there is some planning to be done here.

It would make sense to collect here what kind of information users want to access and list so we can rethink properly the `TransformationChecker` and `Inspector` class.

pomerlef:
I'm closing this pull request, but we will keep track of the request in #23.

PR #165, opened 2016-07-11
Transformation.cpp: There is no chain, the transform must be applied once

oleg-alexandrov:


pointmatcher/Transformation.cpp
```diff
@@ -60,12 +60,22 @@ template struct PointMatcher<double>::Transformation;
...
+
 	DataPoints transformedCloud;
 	for (TransformationsConstIt it = this->begin(); it != this->end(); ++it)
 	{
 		transformedCloud = (*it)->compute(cloud, parameters);
 		swapDataPoints(cloud, transformedCloud);
+		num_iter++;
 	}
+	if (num_iter != 1)
+		throw "Transformations: Error, the transform should have been applied just once.";
```
HannesSommer: This is going to throw an instance of `const char *`. 
I would suggest throwing something more usual, such as `std::runtime_error`. Otherwise it is very unlikely that this text will be visible in case of this exception. Most likely the user is going to see something like 
```
terminate called after throwing an instance of 'char const*'
Aborted
```
instead.

oleg-alexandrov: I did not give this much thought honestly. Here's a new pull request with the exception type you suggest. https://github.com/ethz-asl/libpointmatcher/pull/166

PR #168, opened 2016-07-29
Fix/some small valigrind motivated fixes

HannesSommer:


pomerlef:
Looks good to me. Thanks for calming down valgrind!

HannesSommer:
Thanks for having a look :).

PR #169, opened 2016-08-03
Feature/add time to vtk export

pomerlef:


pointmatcher/IO.cpp
```diff
@@ -1199,28 +1218,99 @@ typename PointMatcher<T>::DataPoints PointMatcherIO<T>::loadVTK(std::istream& is
...
+		{
+			throw runtime_error(string("Missing time field representing nano seconds. Expecting SCALARS with name " + it->first + "_splitTime_nsec in the VTK file."));
 		}
+
+		// Loop through points
+		Int64Matrix timeData(1,pointCount);
+		for(int i=0; i<it->second.sec.cols(); i++)
+		{
+		
+			timeData(0,i) = (((boost::int64_t) it->second.sec(0,i)) << 32) | ((boost::int64_t) it->second.nsec(0,i));
```
HannesSommer: What is the intended data format of timeData and what addTime below expects?  Does it truly expect it in that form (most significant 32 bit seconds and least significant 32 bit nanoseconds)? If yes, I would document that somehow. At least for me this is a big surprise when looking at the code of addTime. 
Why not use 64bit of nanoseconds? That would make computations like difference between timestamps much less complicated. 
Or is it in fact meant this way and the names, sec and nsec, are misleading?

pomerlef: Yes, the idea is to have time expressed as signed int 64 bits nanosec to be compatible with posix_time and to still allow subtraction for delta time.

VTK only support `unsigned_int`, so it is split in two parts when saved:
https://github.com/ethz-asl/libpointmatcher/pull/169/files#diff-65f29f99818f583e6f26c71cfaade3cdR628

I reuse the bit shift conversion from rostime (sec, nsec) structure, but I don't find the source anymore. It might be wrong.

HannesSommer: Assuming the int64 value should be the signed nanoseconds since the epoch higher 32 bit word will not correspond to seconds since epoch in almost any case, simply because e.g. 1 in the upper word will yield a 64bit value of 2^32 and that is quite far away from 10^9, which is the number of nanoseconds in a second. And that makes the identifier `sec` very misleading - I would call it wrong.

pomerlef: humm indeed, I'll rename the parts related to sec and nsec to high32 and low32. I'll also revisit my code generating those timestamp in ros. Searching for the bit shift, I saw that boost has convenient functions to combine sec and nsec.

pomerlef:
I'm moving on with this merge tomorrow.

PR #179, opened 2016-10-31
Public headers get installed again into ${INSTALL_INCLUDE_DIR}/pointmatcher.

HannesSommer:
(Fixed regression introduced with 1066b29d61e1f55abd93c8e3cf.)


pomerlef:
Ouuuu good catch

PR #200, opened 2017-03-29
Fix markdown syntax error

tushar-dadlani:


pomerlef:
That doesn't seem to make a difference...

pomerlef:
Thanks!

Issue #202, opened 2017-04-27
libnabo (master) started to require c++11 for its dependers


HannesSommer:
This is a difficult and rather urgent (currently two master branches do not compile together) issue iff libpointmatcher cannot make that step, yet.

I only see a few affordable options:
1. require c++11 for libpointmatcher and all its dependers, possibly only if the libnabo version found is actually requiring c++11 (the _forward_ option)
2. revert nabo to pre c++11 (the _backward_ option)
3. make libpointmatcher (master) depend officially on a former version of libnabo (e.g. <=1.0.6) (the _pinning_ option)
4. Hide the "third_party/any.h" (optionally) in libnabo from the depender packages (the _mixed_ option). This would allow nabo to depend on c++11 while pointmatcher doesn't.
5. Hide the use of nabo from libpointmatcher's dependers and only require c++11 for libpointmatcher but not its dependers (another _mixed_ option, actually the simples of all, IMO)

The _mixed_ options would satisfying some CI jobs and break this chain of c++11 requirement that otherwise continues for the packages currently depending on libpointmatcher.

@pomerlef , I guess the big question is: How important is it still for libpointmatcher (master) to support ancient compilers?

pomerlef:
libpointmatcher was fully reliant on C++11 at its first release. Then an industrial asked to run it on Windows (Visual Studio), which had a terrible/incomplete support for C++11 at that time (late 2014). It took me almost two months to replace all C++11 functionalities with boost calls.

From what I understood, you're doing the inverse move with libnabo (i.e., replacing boost calls with C++11)? What is the status of libnabo on Mac/Windows/other than Ubuntu?

I've no idea what is the status on Windows right now. For sure, all this back and forth is not optimal... Also, it doesn't make sense to have one on C++11 and not the other one as libpointmatcher wouldn't install on an old platform anyway.

I guess the forward solution is better, but I'm sure there will be other instabilities to handle...

HannesSommer:
Yes, libnabo seems to have started its journey from boost to c++11. 
Good. What about option 5. then? It can also be seen as the friendly version of the froward option and there is almost no difference except that the 
```c++
#include "nabo/nabo.h"
``` 
must be be moved away from `PointMatcher.h` e.g. to `MatchersImpl.h`. This makes it more friendly (and a bit faster to compile against) because it doesn't require the depending package to go through libnabo headers.

[...]

PR #203, opened 2017-04-27
Moved include of nabo to MatchersImpl.h, where it is actually required

HannesSommer:


HannesSommer:
First check is failing as expected because libnabo requires its clients to enable c++11. Why the second is succeeding a mystery to me (but should be unrelated).

HannesSommer:
This should be save (except for cases where clients erroneously rely on pointmatcher to include nabo for their own usage) and useful independent of #202 and #204 . Merging.

PR #206, opened 2017-05-09
Fix compilation with clang

HannesSommer:


pomerlef:
Thanks @HannesSommer !

PR #208, opened 2017-05-12
Remove duplicated code in PointToPlaneErrorMinimizer

davidlandry93:
This is a refactoring PR. I did the refactoring thinking I would need to modify the point to plane error minimizer afterwards, but I ended up not needing it. I think the refactoring is still worth it. In short this PR does the following:

- Removes the duplicated code in `PointToPlaneErrorMinimizer` and `PointToPlaneWithCovErrorMinimizer`. The former is now a parent of the latter, and they share the optimization code. Only the functions that are specific to the covariance remain in `PointToPlaneWithCovErrorMinimizer`.
- Pulls out the `PointToPlaneErrorMinimizer` and `PointToPlaneWithCovErrorMinimizer` classes into their own file. It shortens the size of ErrorMinimizerImpl.cpp and makes the code more readable overall.
- Adds unit tests for the "with cov" versions of the optimizers.


pointmatcher/ErrorMinimizersImpl.h
```diff
@@ -95,30 +95,6 @@ struct ErrorMinimizersImpl
 		virtual T getOverlap() const;
 	};
 
-	struct PointToPlaneErrorMinimizer: public ErrorMinimizer
```
davidlandry93: Is it supported to create error minimizers from outside the library?
I think that moving this class away from this file implies that doing `new ErrorMinimizerImpl<float>::PointToPlaneErrorMinimizer()` does not work anymore. If we still want to support that from outside the library, we could add a typedef inside `ErrorMinimizerImpl<T>` that points to the new class.

HannesSommer:
ok to test

pointmatcher/ErrorMinimizersImpl.h
pomerlef: One of the idea was to allow user to build an ICP solution step by step in their code using the provided building blocks. I doubt that it is used much, but I think it's still a pertinent feature.

utest/ui/ErrorMinimizers.cpp
```diff
@@ -34,14 +34,29 @@ class ErrorMinimizerTest: public IcpHelper
 
 TEST_F(ErrorMinimizerTest, PointToPointErrorMinimizer)
 {
-	setError("PointToPointErrorMinimizer");	
+	setError("PointToPointErrorMinimizer");
```
pomerlef: Those kind of tests are now handled by adding a yaml file in libpointmatcher/examples/data/icp_data/

Here is the info I could find about that in a pull request:

> The idea here is that when a new test is added, or existing behavior is modified, the developer does not have to edit a lot of code. Just data files on disk need to be added or updated (and checked in). For example, say a new yaml file is added, testing a new filter. Running the tests will make the framework iterate over all yaml files, including the new one. The "current" ICP transform for the new yaml file will be written to disk. There was no reference (known true) ICP transform yet to compare with, so the test for the current yaml file will fail. Then copy the newly created "current" ICP transform over the future "reference" (known as correct) ICP transform file. This time when the tests are run they will succeed as the two data files have identical data. Same logic is used if existing behavior is modified.

I should add that in the unit test tutorial at some point...

pointmatcher/ICP.cpp
```diff
@@ -43,6 +43,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "MatchersImpl.h"
 #include "OutlierFiltersImpl.h"
 #include "ErrorMinimizersImpl.h"
+#include "PointToPlaneErrorMinimizer.h"
```
davidlandry93: The PointToPoint part of the job hasn't been done yet but I put the include for all the PointToPlane minimizers

Issue #209, opened 2017-05-22
ICPSequence reprocesses mapPointCloud every registration


lorenwel:
When using an `ICPSequence` every time the compute function is called the `mapPointCloud` gets copied and filtered and the KDTree is rebuilt. 
This slows registration against large point clouds down significantly, e.g. when localizing against a large, static map.
This code was introduced in #57.

Shouldn't this code be moved to the `setMap(...)` function, since this function is required to be called before registering, anyway?
I understand that you might not want `mapPointCloud` to be a filtered map, since we might want to retrieve the map again at a later point. 
In this case an additional member could be added to the `ICPSequence` struct to store the filtered map.
This shouldn't increase memory usage, since in the current state, a filtered copy of the map is created as well. 

Am I missing a reason why we want to reprocess the map every time we register a reading point cloud?
If not and you agree this is an issue, I can open a pull request.

pomerlef:
You're right. The kd-tree shouldn't be rebuilt at every use and filtering should happen outside. 

If you've a pull request to fix it, I would be happy to look at it.

Thanks for reporting!

PR #210, opened 2017-05-26
Fixed rebuilding kdtree every registration

lorenwel:
Addressing #209.

The reference filters are now applied directly to `mapPointCloud` in the `setMap(...)` function.
This means that the filtered map is stored and no copy of the original map is retained. 

The kdtree is now only built when setting the map and not when computing a registration.


lorenwel:
The checks fail because the number of points changes (due to the reference filters being applied directly to the stored map):
```
[ RUN      ] icpTest.icpSequenceTest
/home/jenkins/workspace/libpointmatcher/utest/utest.cpp:269: Failure
Value of: pts0.getNbPoints()
  Actual: 24989
Expected: map.getNbPoints()
Which is: 12617
/home/jenkins/workspace/libpointmatcher/utest/utest.cpp:274: Failure
Value of: pts0.getNbPoints()
  Actual: 24989
Expected: map.getNbPoints()
Which is: 12617
/home/jenkins/workspace/libpointmatcher/utest/utest.cpp:279: Failure
Value of: pts0.getNbPoints()
  Actual: 24989
Expected: map.getNbPoints()
Which is: 12617
[  FAILED  ] icpTest.icpSequenceTest (158 ms)
```
Should we change the test or keep a copy of the original, unfiltered map?

pointmatcher/ICP.cpp
```diff
@@ -470,6 +470,10 @@ bool PointMatcher<T>::ICPSequence::setMap(const DataPoints& inputCloud)
 	// from here reference is express in frame <refMean>
 	// Shortcut to do T_refIn_refMean.inverse() * reference
 	mapPointCloud.features.topRows(dim-1).colwise() -= meanMap.head(dim-1);
+
+  // Apply reference filters
+  this->referenceDataPointsFilters.init();
```
HannesSommer: You use spaces here for indentation but the file is indented with tabs.

HannesSommer:
For the failing unit tests I'd relax to a test `> 0` where appropriate or change the filters such that they do not remove points.

HannesSommer:
Unfortunately this changes the semantics of **public** `getInternalMap`. It could break existing code. Maybe we should rename this function within this PR. E.g. to `getPrefilteredInternalMap`. This way users at least realize the breaking change.

lorenwel:
For the unit test I'd rather use the `>0` condition.
Changing the applied filters might have unwanted side-effects in other unit tests. 

Good point about `getInternalMap`. 
I'll also change the name of `get Map()` then, since it returns the same filtered map, just in global coordinates.

bochen87:
Hi all,

I just tried to apply this patch and I noticed it will break the compilation of the ICP Mapping (https://github.com/ethz-asl/ethzasl_icp_mapping), since the functions are now re-named. So a re-factoring will also need to be applied there.
screenshot from 2017-05-30 12-22-21

pomerlef:
Where are we with that? Was the comment of @bochen87 addressed?

lorenwel:
I created a PR to get things going again https://github.com/ethz-asl/ethzasl_icp_mapping/pull/57

I simply renamed all functions. I'm not familiar with the icp_mapping code so I don't know if this will have any undesired side-effects.

HannesSommer:
What about adding deprecated functions with the old names (getInternalMap and getMap which forward) to make the transition more smooth? It would probably speed up the merging.

An okay deprecation macro can be copied from here: https://github.com/ethz-asl/schweizer_messer/blob/master/sm_deprecation/include/sm/deprecation.h

pomerlef:
I'll trying to clean the PR pending. Where are we with this one?

lorenwel:
I'll add the deprecation warnings if that's fine with everyone.

HannesSommer:
I would be very happy with it :). Best is probably to mention some link to from where the background and the semantics of the change is extractable. E.g. this issue? Or is there a better place?

lorenwel:
I shamelessly copied and adapted the deprecation warnings from schweizer_messer. Let me know if I need to give some reference to that. 
I also just copied the copyright header from an existing file. Let me know if I need to do something different.

HannesSommer:
Great! You may just copy this from SM (I wrote these macros). I believe the copied header is also fine.

Maybe we should mention that these were possibly breaking changes that are already applied. Maybe add something like 'These changes may already have altered your program and action may be required!' to the messages. (Normally people would expect deprecated function to at least act as before till they get removed).

HannesSommer:
I though I had written this already: Well done @lorenwel  ! @pomerlef  I'd recommend merging. Actually, a rebase could be nice to get rid of the many merge commits in between.

pomerlef:
Thanks @lorenwel for the fix and thanks @HannesSommer for the coordination!

Issue #215, opened 2017-08-07
Reflection Matrix in PointToPointCovErrorMinimizer


pierrefranklin:
Taken directly from the code in PointToPointErrorMinimizer::compute(...):

```
// It is possible to get a reflection instead of a rotation. In this case, we
// take the second best solution, guaranteed to be a rotation. For more details,
// read the tech report: "Least-Squares Rigid Motion Using SVD", Olga Sorkine
// http://igl.ethz.ch/projects/ARAP/svd_rot.pdf
```

The code around this (lines 220 to 234) is missing from the PointToPointCovErrorMinimizer (lines 333 to 340) in ErrorMinimizersImpl.cpp.

pomerlef:
Good catch. That gives me another argument to start refactoring `PointToPoint` functions as it happen with `PointToPlane`.

I'll try to do that sooner that later.

artivis:
Any chance to see those lines simply copied while waiting for the refactoring :) ??

pomerlef:
Yes, it would make sense. I'm not installed to check that on my side.

You can send a pull request if you feel comfortable with it. Otherwise, @davidlandry93 could you have a look?

artivis:
see PR #217

pomerlef:
Merged, thanks @artivis!

PR #217, opened 2017-08-14
handle reflexion in PointToPointWithCovErrorMinimizer

artivis:
Answering issue #215 .


pomerlef:
ok to test

pomerlef:
@HannesSommer the error seems to be related to the system installation:
```
CMake Error:
  Error evaluating generator expression:

    $<TARGET_FILE:nabo>

  No target "nabo"
```

HannesSommer:
@pomerlef , that's a bit of a nasty one. `cmake` is finding random (old) versions of `libnaboConfig.cmake` on Jenkins. I'm on it.

HannesSommer:
Okay, PR #218 should solve the build issue discovered here together with the Jenkins changes I already applied.

pomerlef:
Awesome, thanks a lot, @HannesSommer. Do we need to wait for the PR in libnabo to be accepted too?

HannesSommer:
No, this works without that PR for libnabo, because it is now finding properly the installed (vs. build) version of the `libnaboConfig.cmake` file. That PR is for the build version, which libpointmatcher was using before (most of the times).

PR #218, opened 2017-08-15
Fixed usage of the LIBNABO_INSTALL_DIR variable

HannesSommer:
This is how one can amend the prefix list for find_package.

This should help compiling #217 .


HannesSommer:
There two other bugs kicking in and contributing to the problem: one in our Jenkins scripts and one in libnabo :(. ' Hope to fix both soon.

pomerlef:
seems good to me

Issue #228, opened 2017-11-23
Libpointmatcher segfaults with latest libnabo changes


bochen87:
See details at https://github.com/ethz-asl/libnabo/issues/81

Summary:
Probably segfaults due to libnabo now returning -1 instead of 0 for an invalid index here: https://github.com/ethz-asl/libpointmatcher/blob/75044815d40ff934fe0bb7e05ed8bbf18c06493b/pointmatcher/OutlierFiltersImpl.cpp#L253

HannesSommer:
Thanks for "moving" here.
 
Actually the invalid index is NNS::InvalidIndex (-1 for signed indices) : https://github.com/ethz-asl/libnabo/blob/master/nabo/nabo.h#L343. 

I'll try to propose a fix.

bochen87:
fixed with #230

PR #229, opened 2017-11-23
Avoid using invalid matches

HannesSommer:
Since https://github.com/ethz-asl/libnabo/pull/79 the invalid index is no longer `0`. 

This revealed that libpointmatcher was actually using (IMO) invalid matches as if they were matches with the first entry. (see #228 ).

Currently this is only a start fixing this: It affects more than just this line. And my fix is probably not the best.

@bochen87, if you are still motivated to help: could you test this branch together with libnabo master?


HannesSommer:
Sorry @bochen87 , I forgot to set the weight with my first commit. Now it should be okay to test.

bochen87:
Wow that was a fast fix! I will try out tomorrow!

PR #232, opened 2017-12-11
Changed pass by const value to pass by const reference to remove the extra copy.

BenBallard:
When making these error minimizer objects a copy of the source data points is made.  If the number of source points is large and then making this copy can be really expensive.  Since these are passed by const value they can be passed by const reference to remove the copy.


HannesSommer:
ok to test

PR #235, opened 2018-01-14
remove `grep` dependency, parse version with CMake

svenevs:
- Parse version using CMake rather than `grep`, makes it easier for Windows people.
- Removed `grep` install step from windows instructions.
- Renamed `PROJECT_VERSION` to `POINTMATCHER_PROJECT_VERSION`.
- Deal with CMP0048 by getting the project version first before `project(libpointmatcher)`.

I grepped around for `grep` and didn't see any other place, but maybe it is used for something more?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,27 +1,38 @@
 cmake_minimum_required(VERSION 2.6)
 
-include (CheckSymbolExists)
+include(CheckSymbolExists)
 
-project(libpointmatcher)
+#========================
+# Project details / setup
+#========================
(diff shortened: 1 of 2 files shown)


pomerlef:
ok to test

pomerlef:
Thanks a lot. I wanted to do that since a while!

svenevs:
Sure thing!  I saw the `grep` thing and decided to do a quick search online, fully expecting it to be really complicated.  It was a pleasant surprise how simple it was :smile:

Issue #234, opened 2018-01-12
[question] externally coordinating normals with ICP


svenevs:
Hello,

Please feel free to close without response, as this is not an issue with the library, it's me not seeing the whole picture.

1. I already have normals computed, so I should not be writing a `DataFilter`.  I should be adding a [descriptor](http://libpointmatcher.readthedocs.io/en/latest/Pointclouds/#descriptor-labels) when I create `DataPoint` objects, correct?

2. The descriptor for (1) **must** be named `"normals"`, emphasis: **`s`** at the end (if so, should I PR to add an `s` on the docs?):

    https://github.com/ethz-asl/libpointmatcher/blob/59164e766d9964255031a6b1415e4c9af558abaf/pointmatcher/ErrorMinimizers/PointToPlane.cpp#L115-L116

3. So assuming I successfully create (1) with `normals` labelled correctly, this is enough for the actual implementation to use it right?  Or is there another step to making sure these actually get used?

4. Since I'm externally specifying normals, but do need to remove NaN, I need to pre-filter the buffers I use as input positions / normals?  AKA the NaN `DataFilter` would only filter positions, but I've got data coming in where if the position was NaN, the normal is set to NaN.
    - Asked another way: if I want to externally supply normals, then I really cannot use any existing `DataFilter` objects, since whatever would get filtered here would mean that my positions dimensions and normals dimensions no longer agree?

I'm very new to this library, and most confused about (4), since it seems like that would mean the NormalsFilter must **always** be specified last, but this doesn't seem to be true, nor would it make sense because you would get very different results for the normals if you severely downsample.

Thank you for any thoughts / clarifications!

pomerlef:
1. yes
1. yes, yes
1. yes, the manual generation of DataPoints is not well documented (lack of time). Your best documentation is the unit tests. The method `addDescriptor()` is the easiest if your `DataPoints` already exists:

   https://github.com/ethz-asl/libpointmatcher/blob/master/utest/ui/DataPoints.cpp#L323-L324

   Otherwise, a constructor with descriptor exists too: 

   https://github.com/ethz-asl/libpointmatcher/blob/master/utest/ui/DataPoints.cpp#L57

1. There are many checks to ensure that the number of points in the feature and descriptor matrix are the same. If a filter remove a point (either based on a feature or on a descriptor), both matrices are resized.

svenevs:
[...]

PR #236, opened 2018-01-16
[WIP] manual data and descriptors example

svenevs:
Here's what I've got so far on manually specifying positions / normals using `Eigen::Map`.  I used a `float4` data type, but maybe this adds more confusion than it does help?  I don't find it counter-intuitive, but I hail from a CUDA background...so of course it is natural.

Opening the PR now for feedback, we can just squash the commits when its ready / if it is desired.

- `sin*cos` is kind of a bad choice, if you do [0, 2pi] it won't converge.  just did 0.5 pi to 3pi / 2.  [the translation of only z is also important](https://github.com/svenevs/libpointmatcher/blob/2926febba6eb1b0080f6f53daca88338a220d30f/examples/manual_data_and_descriptors.cpp#L248-L256).  but i think for the purposes of this demo it is sufficient
- normals are working, if you remove the `topRows(3)` and just use the map you'll get an exception from libpointmatcher
    - obviously, the normals are quite naive

Bad alignment [0, 2pi]:

manual

Good enough alignment:

manual_fixed

- Purple: reference frame
- Teal: reading with transform applied
- Blue: reading frame

**TODO**:

- [x] features are problematic, segfault at end of program (not sure where from yet).  double free or something probably.  double free from reusing the NaN filter, created two -- one for reading and one for reference.
- [ ] need writeup on the tutorial instead of [linking from index.md](https://github.com/svenevs/libpointmatcher/blob/2926febba6eb1b0080f6f53daca88338a220d30f/doc/index.md#advanced) {revert that}
- [x] need to fix the NaN filter for descriptors (see [comment here](https://github.com/ethz-asl/libpointmatcher/issues/234#issuecomment-357778015))
    - [x] added to description, see [note at end here](https://github.com/svenevs/libpointmatcher/blob/manual_data_and_descriptors/doc/Datafilters.md#remove-nan-filter-)
        - [x] decision from you all: the datafilters markdown had whitespace trimmed from my text editor.  for the descriptions to show up they were, this required trailing whitespace.  I fixed this with newlines, but it was ugly.  So I made them all bullet lists.  See [the bullet version](https://github.com/svenevs/libpointmatcher/blob/manual_data_and_descriptors/doc/Datafilters.md#surface-normal-filter-) vs [the original version](https://github.com/ethz-asl/libpointmatcher/blob/master/doc/Datafilters.md#surface-normal-filter-).  I can find a way to un-do this easily enough.  But I'm not sure that relying on trailing whitespace is valid markdown, and it may break in the future.  I couldn't figure out how to build the RTD docs, so don't know what that looks like.  Just let me know what you want.
    - [x] added explicit nan test, test 1 validates existing behavior, test 2 validates NaN only in normals makes both get skipped
    - [ ] [**Confusion: why don't I need to check that `.col(i)` is a valid index for the descriptors**?](https://github.com/svenevs/libpointmatcher/blob/a93273ff385d946cacfc24c81b13719040432472/pointmatcher/DataPointsFiltersImpl.cpp#L96-L98)

### DIFF
--- a/doc/Datafilters.md
+++ b/doc/Datafilters.md
@@ -61,8 +61,11 @@ The following examples are drawn from the apartment dataset available for [downl
 Points can be excluded from a rectangular bounding region by using this filter.  The box dimensions are specified by defining the maximum and minimum coordinate values in the x,y,z directions.
 
 __Required descriptors:__ none
+
 __Output descriptor:__ none
+
 __Sensor assumed to be at the origin:__ no
+
 __Impact on the number of points:__ reduces number of points
 
 
@@ -95,9 +98,14 @@ A number of filters are used to reduce the number of points in a cloud by random
 
 Points are only considered for rejection if they exceed a density threshold, otherwise they are preserved.  The single parameter of this filter sets the maximum density that should be obtained in the output point cloud.  Points are randomly rejected such that this maximum density is obtained as closely as possible.
 
-__Required descriptors:__ `densities` (see SurfaceNormalDataPointsFilter and SamplingSurfaceNormalDataPointsFilter)
+__Required descriptors:__
+
+- `densities` (see SurfaceNormalDataPointsFilter and SamplingSurfaceNormalDataPointsFilter)
+
 __Output descriptor:__ none
+
 __Sensor assumed to be at the origin:__ no
+
 __Impact on the number of points:__ reduces number of points
 
 |Parameter  |Description  |Default value    |Allowable range|
@@ -116,8 +124,11 @@ In the following example we observe the effect of the maximum density filter on 
 These filters remove points which lie beyond a threshold distance from the coordinate center.  Points are kept if their distance from the center **greater than** the threshold.  The distance threshold can be defined on the x,y, and z axes or can be a radial distance from the center.
 
 __Required descriptors:__ none
+
 __Output descriptor:__ none
+
 __Sensor assumed to be at the origin:__ no
+
 __Impact on the number of points:__ reduces number of points
 
 |Parameter  |Description  |Default value    |Allowable range|
@@ -137,8 +148,11 @@ In the following example, a maximum distance threshold of 1m is applied radially
 These filters remove points which lie beyond a threshold distance from the coordinate center.  Points are kept if their distance from the center **smaller than** the threshold.  The distance threshold can be defined on the x,y, and z axes or can be a radial distance from the center.
 
 __Required descriptors:__ none
+
 __Output descriptor:__ none
+
 __Sensor assumed to be at the origin:__ no
+
 __Impact on the number of points:__ reduces number of points
 
 |Parameter  |Description  |Default value    |Allowable range|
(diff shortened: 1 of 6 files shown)


pomerlef:
ok to test

pomerlef:

- I'm fine with the bullet version of the doc.
- Would you mind adding a separate .md file to explain roughly/quickly your code and add the screenshots that you put in this PR?

svenevs:
Sure, I should be able to do that within the next couple of days.  Do you want the picture of the bad alignment too?  Or just the good one?

pomerlef:
As you see fit. Try to keep it simple as it is about building your own `DataPoints` using external inputs and not about tuning an ICP solution for a specific case.

svenevs:
As a followup, is the `w` component always required for `features`?  I'm gonna try and test that out and just push back labels `X`, `Y`, and `Z`, but if it works, does that have negative consequences elsewhere?

If it works, it would make sense to include that in the tutorial.  Then it would show multiple options

1. If you already have four floats for positions.
2. How to take four floats for normals (assuming XYZW) and still use it with libpointmatcher.
3. If you only have three floats for positions.

My schedule for tonight bodes well to test this / finalize the writeup :smile:

pomerlef:
tl;dr 
Yes, the `pad` value is necessary.

The fourth entry is the scale (in our case less pertinent, but it is for computer vision). The transformation matrix is 4x4 so it needs to be multiplied by a 4x1 to work.

Instead of adding the one every time we need to do an operation, we added in the `DataPoints` directly. Also, Eigen is optimized for four operation in parallel (if I recall well).

pomerlef:
Is this patch finished? I lost track of the progress.

svenevs:
No sorry :/ I still need to add the explanation in a markdown document. I kind of lost track of this... But should be able to do it this weekend.

pomerlef:
Alright, no problem. I'll merge large refactoring of `DataPointfFilters` today. Just make sure to have your branch up-to-date this weekend.

svenevs:
Going through and closing stale pull requests of mine.  Looking more closely at this tutorial, I don't think it adds much value.  It's also using `_mm_malloc`, which isn't a valid cross-platform solution.  In my opinion, the user who wants to share external data with libpointmatcher falls into one of two categories:

1. They understand the implications of sharing data with Eigen, including memory alignment concerns.  They already know how to use `Eigen::Map`, and this tutorial adds no value.
2. They don't know how to share data with Eigen.  This tutorial is too advanced, doesn't give nearly enough context, and they'll get lost in the boilerplate code to just set the problem up.  What's this magic number `16`?  Why do you have to align data?  All that stuff ...

I don't have time to write a good enough tutorial for (2).  I'm attaching the sample code here because I'm deleting that fork and I don't know what happens to the diff / if it remains viewable, in case you feel you would like to re-work it on your own.  If you were to do so, I'd suggest making more forward declarations so that the important part of the tutorial (showing how to `Map`, line `260` - `368` are the only important parts in my opinion) is higher in the file, as well as probably just using [Eigen's aligned allocator](https://eigen.tuxfamily.org/dox/group__TopicStlContainers.html) and doing something with `std::vector<Eigen::Vector4f>` or something.  Then you can get around the fact that `std::aligned_alloc` isn't available until C++17 :(

Sorry this was open for so long with no progress only to be closed.

[`manual_data_and_descriptors.cpp.txt` (added `.txt` extension so GitHub lets me attach it)](https://github.com/ethz-asl/libpointmatcher/files/3562132/manual_data_and_descriptors.cpp.txt)

PR #237, opened 2018-01-16
fix two typos

svenevs:
One of these was from my earlier PR switching from `grep` to CMake parsing, sorry!


HannesSommer:
ok to test

PR #243, opened 2018-05-14
Fix some errors in DataFilters documentation

MathLabu:
- Voxel sizes parameters of `VoxelGridDataPointsFilter` previously in _]-inf;+inf[_ range, so change the documentation to be in _]0;+inf[_ correct range
- Dead link in SurfaceNormalFilter documentation has been removed


pomerlef:
add to whitelist

PR #244, opened 2018-05-21
Fix/datafilters : clean Data Filters before reorganizing them

MathLabu:
- add `const` qualifiers where needed  
- change from postfix to prefix (`i++` to `++i`)
- set variables for loop stop condition
- adapt brace style to convention


pomerlef:
Nice cleanup, thanks for the effort @MathLabu !

PR #246, opened 2018-05-22
Refactoring of the Data Filters

MathLabu:
I refactored all the Data Filters (DF) previously in `DataPointsFiltersImpl.h` and `DataPointsFiltersImpl.cpp` as follow:

- Create a new directory named `DataPointsFilters/` containing now the DFs
- Separate each DF in a specific file (.h and .cpp):  for instance the `IdentityDataPointsFilter` is now declared in `DataPointsFilters/Identity.h` and `DataPointsFilters/Identity.cpp`
- Each DF is now templated on `T` (`float` or `double`) and inherit directly of `PointMatcher<T>::DataFilter`
- Still API compatible due to a _trick_ in `DataPointsFiltersImpl.h` where each DF is typedef-ed as follow: 
```cpp
template<typename T>
struct DataPointsFiltersImpl {
    typedef ::IdentityDataPointsFilter<T> IdentityDataPointsFilter;
};
```
so, registration of DF in the registrar works without any change and the headers are included at only one place (`DataPointsFiltersImpl.h`). However, it's possible to directly change the `Registry.cpp` file not to be dependent of `DataPointsFiltersImpl` class anymore
```cpp
ADD_TO_REGISTRAR_NO_PARAM(DataPointsFilter, IdentityDataPointsFilter, typename DataPointsFiltersImpl<T>::IdentityDataPointsFilter)
```
- Add an `utils.h` header containing often used functions by the DFs (in a namespace `utils`, such as `computeNormals(...)`, etc.)
- Add minor improvements in `Elipsoids`, `SamplingSurfaceNormal` and `Gestalt` DFs by introducing move semantic where needed.

I'll update the tutorial to adapt it to these changes too! Hope it'll be helpful!


pomerlef:
Alright, thanks for the thorough cleanup!

Since it's a large PR, I'll let it open for a day just in case someone has something to say.

PR #250, opened 2018-05-31
Add a new datafilter based on octree

MathLabu:
I developed an [Octree](https://en.wikipedia.org/wiki/Octree) structure to be used by the library:
- the current implementation ensures to have either 0 or 8 children for each node
- only leaf nodes contain data
- data contained by the octree are currently the indexes of the points from the `DataPoint` structure
- processing are applied by calling a callback for each node in a depth-first search ([DFS](https://en.wikipedia.org/wiki/Depth-first_search))
- parallel build can be enabled (but only the first level is parallelized)
- the octree can be build given two criterion: number of data by leaf node, or size of the leaf node

**Remark 1:** currently the space is decomposed from the Cartesian coordinates (x,y,z), but there should be no limitation to use other dimensions (such as normal coordinates for instance).

**Remark 2:** currently the octree only stands for 3D point point-cloud, a Quadtree (_same principle but for 2D space_) could easily be developed from this implementation.

-----
With this new structure, I developed a `DataPointsFilter` named `OctreeGridDataPointsFilter`:
- build an octree spatial representation of the point cloud
- apply a process to sample in each leaf node

Three sampling methods have been developed:
- take the first point of each node (`FirstPtsSampler`)
- take a random point of each node (_quite similar to the previous one since the point cloud is not supposed to be ordered_) (`RandomPtsSampler`)
- compute centroid of each node (_more precise but more costly_) (`CentroidSampler`)  

Theoretically, any process can be applied to the point cloud (sampling, feature enhancement, filtering, etc.) since the octree give a spatial representation of the point cloud (_ex: we could estimate the normal of each leaf_).

-----
To validate the filter, I also write a `utest`.

-----
**Remark 3:** since the octree give an efficient spatial representation of space (empty space is not decomposed), the voxel representation should be deprecated.

I am pretty sure there is a lot of thing to improve this octree (I'm open to any suggestion) but it does the work for the datafilters.


pomerlef:
Thanks @MathLabu !

pointmatcher/DataPointsFilters/utils/octree.h
```diff
@@ -0,0 +1,167 @@
...
+public:
+	using PM = PointMatcher<T>;
+	using DP = typename PM::DataPoints; /**/
+	using Id = typename DP::Index; /**/
+	
+	using Data = typename DP::Index; /**/
+	using DataContainer = std::vector<Data>;
+
+private:
+	struct XYZ
```
stephanemagnenat: Isn't that duplicating some functions already provided by Eigen by `Eigen::Matrix<T,3,1>`? It would be nice not to duplicate such basic features.

pointmatcher/DataPointsFilters/utils/octree.hpp
```diff
@@ -0,0 +1,609 @@
...
+		//nullify childs
+		for(size_t i=0; i<8; ++i)
+			octants[i]= nullptr;
+		//Copy data
+		data.insert(data.end(), o.data.begin(), o.data.end());
+	}
+	else //Node case
+	{
+		//Create each child recursively
+  	for(size_t i=0; i<8;++i)
```
stephanemagnenat: Weird indentation

pointmatcher/DataPointsFilters/utils/octree.h
MathLabu: Yes indeed it's kinda a duplicate, but for efficient memory representation I needed a minimal struct for representation a triplet of coordinate. 

However, I tested a version a the octree with `Eigen::Matrix<T,3,1>` instead of the `XYZ` struct. A quick benchmark tends to suggest that the build of the octree with the latter is quicker than with `Eigen` representation, which is `1.89` times slower. 

Any idea how to optimize the use of the `Eigen` matrix (where the XYZ is used) to speedup the build ?

utest/ui/DataFilters.cpp
```diff
@@ -442,6 +440,94 @@ TEST_F(DataFilterTest, MaxPointCountDataPointsFilter)
...
+	const unsigned int nbPts = 60000;
+	const DP cloud = generateRandomDataPoints(nbPts);	
+	params = PM::Parameters(); 
+	
+//CASE (1): 1/pts by octants + validate parallel build
+// the number of point should not change
+// the sampling methods should not change anything
+	PM::DataPointsFilter* octreeFilter;
+	
+	const vector<int> samplingMethods = list_of (0) (1) (2);
```
HannesSommer: `samplingMethods{0, 1, 2}` initialization would be more standard and shorter.

utest/ui/DataFilters.cpp
```diff
@@ -442,6 +440,94 @@ TEST_F(DataFilterTest, MaxPointCountDataPointsFilter)
...
+	const vector<bool> parallels = list_of (true) (false);
```
HannesSommer: same as above.

utest/ui/DataFilters.cpp
```diff
@@ -442,6 +440,94 @@ TEST_F(DataFilterTest, MaxPointCountDataPointsFilter)
...
+			EXPECT_EQ(cloud.getNbPoints(), filteredCloud.getNbPoints());
+			EXPECT_EQ(cloud.getDescriptorDim(), filteredCloud.getDescriptorDim());
+			EXPECT_EQ(cloud.getTimeDim(), filteredCloud.getTimeDim());
+		
+			EXPECT_EQ(filteredCloud.getNbPoints(), nbPts);
+		}
+	}
+	
+//CASE (2): 200/pts by octants	
+	for(const auto& meth : samplingMethods)
```
HannesSommer: Couldn't you just make another loop around the nested loop above over "1", "200" for maxPointByNode ? What else is specific here? That way you would test more combinations, which is probably good.

utest/ui/DataFilters.cpp
```diff
@@ -442,6 +440,94 @@ TEST_F(DataFilterTest, MaxPointCountDataPointsFilter)
...
+		EXPECT_GT(cloud.getNbPoints(), filteredCloud.getNbPoints());
+	
+		//Validate transformation
+		icp.readingDataPointsFilters.clear();
+		addFilter("OctreeGridDataPointsFilter", params);
+		//validate2dTransformation();
+		validate3dTransformation();
+	}
+
+//CASE (3): octants max size 10cm
```
HannesSommer: Same as above. This is again too much duplication!
If you merge them all you can also inline `samplingMethods` like I suggested for `parallels` above.

utest/ui/DataFilters.cpp
```diff
@@ -442,6 +440,94 @@ TEST_F(DataFilterTest, MaxPointCountDataPointsFilter)
...
+	
+//CASE (1): 1/pts by octants + validate parallel build
+// the number of point should not change
+// the sampling methods should not change anything
+	PM::DataPointsFilter* octreeFilter;
+	
+	const vector<int> samplingMethods = list_of (0) (1) (2);
+	const vector<bool> parallels = list_of (true) (false);
+	
+	for(const auto& par : parallels)
```
HannesSommer: Consider manual "inlining" `parallels` (makes it more readable IMO):

```c++
for(const bool doParallel : {false, true})
...
```

pointmatcher/DataPointsFilters/utils/octree.hpp
```diff
@@ -0,0 +1,609 @@
...
+		octants[i]->depth = this->depth+1;
+		ret = ret and octants[i]->build(pts, std::move(sDatas[i]), std::move(boxes[i]), maxDataByNode);		
+		//Assign parent
+		octants[i]->parent = this;
+	}
+
+	return (!isLeaf() and ret);
+}
+
+#include <thread>
```
HannesSommer: This include should go to the top.

pointmatcher/DataPointsFilters/utils/octree.h
```diff
@@ -0,0 +1,167 @@
...
+	Octree* parent;
+	Octree* octants[8];
+	
+	/******************************************************
+	 *	Octants id are assigned as their position 
+	 *   on a hamming cube (+ greater than center, - lower than center)
+	 *
+	 *	  	0	1	2	3	4	5	6	7
+	 * 	x:	-	+	-	+	-	+	-	+
+	 * 	z:	-	-	+	+	-	-	+	+
```
HannesSommer: Is there a special reason for this counterintuitive ordering? (z before y). If so maybe state it?

pointmatcher/DataPointsFilters/utils/octree.hpp
```diff
@@ -0,0 +1,609 @@
...
+		boxes[i].radius = half_radius;
+		boxes[i].center = this->bb.center + offset;
+	}
+	
+	//For each child build recursively	
+	std::vector<std::thread> threads;
+	bool ret = true;
+	for(size_t i=0; i<8; ++i)
+	{
+		threads.push_back( std::thread( [maxSizeByNode, i, &pts, &sDatas, &boxes, this](){
```
HannesSommer: `std::async` with the `std::launch::async` policy (!) would have been better suited IMO (leaving more decisions to the implementation (e.g. actual number of threads to optimize hardware utilization).

pointmatcher/DataPointsFilters/utils/octree.h
MathLabu: You are right, I was missing the optimization flag. With `-O3` the build time is the same for both structure, I'll switch to `Eigen::Matrix<T,3,1>` then. Thanks for pointing that out.

pointmatcher/DataPointsFilters/utils/octree.h
MathLabu: In fact, I build the indexes as their position on a hamming cube, considering the axes as the same for 3D coordinates (direct system with the `000` point as origin). So the z coordinate is the 2nd bit in the index.

hamming_cube

If you think there is some drawback to use this notation, please tell me so I can change it.

PR #254, opened 2018-06-01
Add a Quadree struct to manage 2D point

### LINKED ISSUE #251: Quadtree structure to handle 2D data ?
Currently only 3D data are supported by the octree in the library. 

Would be interesting to develop a [Quadtree](https://en.wikipedia.org/wiki/Quadtree) for the 2D data based on the code of the octree.

MathLabu:
(fixes #251)

I added a [Quadtree](https://en.wikipedia.org/wiki/Quadtree) for the 2D data based on the code of the octree (see #250 for more details).

-----
Plus, I updated the`OctreeGridDataPointFilter` so it can handle 3D as well as 2D point cloud. To do so, I had to allow the `Sampler` to be applied on the Octree as well as the Quadtree by templating the `operator()`. However, it's transparent to the user.

-----
I updated the `utest` that now also test 2D transformation.

-----
The choice to develop an Quadtree separately of the Octree can be justify by the following:
- It's the responsibility of the user to choose the correct data structure adapted to his needs
- Templatize the Octree could be done, but imply that we have to specialize some processes for a 3D data or a 2D data (ex: the index calculation, the offset table, etc.)
- Having a different class for each representation let more possibility to optimize each one

The contrary can be justified too: better maintainability, no code duplication, etc. 

-----
I also correct to typo/indentation as reviewed in #250, renamed the struct `XYZ` into `Point` in the Octree and the Quadtree to be homogeneous and made the interfaces of `Point` and `BoundingBox` public. 

----
As for my previous PR, I am open to any suggestion to improve the Octree and the Quadtree.

### DIFF
--- a/pointmatcher/DataPointsFilters/OctreeGrid.cpp
+++ b/pointmatcher/DataPointsFilters/OctreeGrid.cpp
@@ -57,7 +57,7 @@ OctreeGridDataPointsFilter<T>::FirstPtsSampler::FirstPtsSampler(DataPoints& dp):
 }
 
 template <typename T>
-template<template<typename> typename Tree>
+template<template<typename> class Tree>
 bool OctreeGridDataPointsFilter<T>::FirstPtsSampler::operator()(Tree<T>& oc)
 {
 	if(oc.isLeaf() and not oc.isEmpty())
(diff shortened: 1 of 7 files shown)


MathLabu:
I consider taking into account the reviews from the PR #250 !

I'll reopen and update this PR once the other issues are resolved.

PR #255, opened 2018-06-04
Major revision of the Octree class

### LINKED ISSUE #251: Quadtree structure to handle 2D data ?

MathLabu:
I took into account the reviews from the previous PR about the octree #250 :
- The octree is now templated on the dimension (`dim=3` for the [Octree](https://en.wikipedia.org/wiki/Octree) (3D) case; `dim=2` for the [Quadtree](https://en.wikipedia.org/wiki/Quadtree) (2D) case)
- No more code duplication. Only one `build` function that handle parallel build and with merged stop criterion
- The struct `XYZ` has been replace by `Eigen::Matrix<T,dim,1>` to avoid duplication
- The use of macro has been replaced by functions (ex: `toData`, overload of `idx`)
- `std::thread` has been replaced by `std::async` with `std::launch::async` policy

The new implementation natively support the 2D case with the Quadtree (i.e. a `Quadtree` is an `Octree_<T,2>`).   (fixes #251) 

I created the following aliases not to break the existing API:
```c++
template<typename T> using Quadtree = Octree_<T,2>;
template<typename T> using Octree = Octree_<T,3>;
```

----
The unit test has been rewrite too, in a more compact way and with 2D test case for the Quadtree.

---
I hope this version is cleaner than the previous one. Still open to suggestion, ~in particularly for the choice of the 3D indexes on an Hamming cube (as discussed in #250)~.

**Edit:** the indexes on the Hamming cube had been removed to have a more natural order.

### DIFF
--- a/pointmatcher/DataPointsFilters/OctreeGrid.cpp
+++ b/pointmatcher/DataPointsFilters/OctreeGrid.cpp
@@ -171,9 +171,11 @@ bool OctreeGridDataPointsFilter<T>::CentroidSampler::operator()(Octree<T>& oc)
 		//retrieve index from lookup table if sampling in already switched element
 		if(std::size_t(d)<idx)
 			j = mapidx[d];
-			
+		
+		T acc[featDim-1] = {0.};
+		
 		//We sum all the data in the first data
-		for(std::size_t id=1;id<nbData;++id)
+		for(std::size_t id=0;id<nbData;++id)
 		{
 			//get current idx
 			const auto& curId = (*data)[id];
@@ -184,7 +186,7 @@ bool OctreeGridDataPointsFilter<T>::CentroidSampler::operator()(Octree<T>& oc)
 				i = mapidx[curId];
 			
 			for (int f = 0; f < (featDim - 1); ++f)
-				pts.features(f,j) += pts.features(f,i);
+				acc[f] += pts.features(f,i);
 			
 			if (pts.descriptors.cols() > 0)
 				for (int d = 0; d < descDim; ++d)
@@ -197,7 +199,7 @@ bool OctreeGridDataPointsFilter<T>::CentroidSampler::operator()(Octree<T>& oc)
 		
 		// Normalize sums to get centroid (average)
 		for (int f = 0; f < (featDim - 1); ++f)
-			pts.features(f,j) /= T(nbData);
+			pts.features(f,j) = acc[f]/T(nbData);
 		
 		if (pts.descriptors.cols() > 0)
 			for (int d = 0; d < descDim; ++d)
@@ -224,20 +226,17 @@ template <typename T>
 OctreeGridDataPointsFilter<T>::OctreeGridDataPointsFilter(const Parameters& params) :
 	PointMatcher<T>::DataPointsFilter("OctreeGridDataPointsFilter", 
 		OctreeGridDataPointsFilter::availableParameters(), params),
-	parallel_build{Parametrizable::get<bool>("buildParallel")},
+	buildParallel{Parametrizable::get<bool>("buildParallel")},
 	maxPointByNode{Parametrizable::get<std::size_t>("maxPointByNode")},
 	maxSizeByNode{Parametrizable::get<T>("maxSizeByNode")}
 {
 	try 
 	{
-		const int bm = Parametrizable::get<int>("buildMethod");
-		buildMethod = BuildMethod(bm);
 		const int sm = Parametrizable::get<int>("samplingMethod");
 		samplingMethod = SamplingMethod(sm);
 	}
 	catch (const InvalidParameter& e) 
 	{
-		buildMethod = BuildMethod::MAX_POINT;
 		samplingMethod = SamplingMethod::FIRST_PTS;
 	}
 }
@@ -257,20 +256,7 @@ void OctreeGridDataPointsFilter<T>::inPlaceFilter(DataPoints& cloud)
 	assert(cloud.features.rows() == 4); //3D points only
 	
 	Octree<T> oc{};
-	
-	switch(buildMethod) 
-	{
-		case BuildMethod::MAX_POINT:
-		{
-			oc.build(cloud, maxPointByNode, parallel_build);
-			break;
-		}
-		case BuildMethod::MAX_SIZE:
-		{
-			oc.build(cloud, maxSizeByNode, parallel_build);
-			break;
-		}
-	}
+	oc.build(cloud, maxPointByNode, maxSizeByNode, buildParallel);
 	
 	switch(samplingMethod)
 	{
--- a/pointmatcher/DataPointsFilters/OctreeGrid.cpp
+++ b/pointmatcher/DataPointsFilters/OctreeGrid.cpp
@@ -57,7 +57,8 @@ OctreeGridDataPointsFilter<T>::FirstPtsSampler::FirstPtsSampler(DataPoints& dp)
 }
 
 template <typename T>
-bool OctreeGridDataPointsFilter<T>::FirstPtsSampler::operator()(Octree<T>& oc)
+template<std::size_t dim>
+bool OctreeGridDataPointsFilter<T>::FirstPtsSampler::operator()(Octree_<T,dim>& oc)
 {
 	if(oc.isLeaf() and not oc.isEmpty())
 	{			
@@ -107,7 +108,8 @@ OctreeGridDataPointsFilter<T>::RandomPtsSampler::RandomPtsSampler(
 	std::srand(seed);
 }
 template<typename T>
-bool OctreeGridDataPointsFilter<T>::RandomPtsSampler::operator()(Octree<T>& oc)
+template<std::size_t dim>
+bool OctreeGridDataPointsFilter<T>::RandomPtsSampler::operator()(Octree_<T,dim>& oc)
 {
 	if(oc.isLeaf() and not oc.isEmpty())
 	{			
@@ -154,7 +156,8 @@ OctreeGridDataPointsFilter<T>::CentroidSampler::CentroidSampler(DataPoints& dp)
 }
 	
 template<typename T>
-bool OctreeGridDataPointsFilter<T>::CentroidSampler::operator()(Octree<T>& oc)
+template<std::size_t dim>
+bool OctreeGridDataPointsFilter<T>::CentroidSampler::operator()(Octree_<T,dim>& oc)
 {
 	if(oc.isLeaf() and not oc.isEmpty())
 	{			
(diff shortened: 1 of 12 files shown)


pointmatcher/DataPointsFilters/utils/utils.h
```diff
@@ -41,6 +41,18 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+template<class T>
+inline constexpr T pow(const T base, const std::size_t exponent)
+{
+    // (parentheses not required in next line)
+    return (exponent == 0)     ? 1 :
+           (exponent % 2 == 0) ? pow(base, exponent/2)*pow(base, exponent/2) :
+           base * pow(base, (exponent-1)/2) * pow(base, (exponent-1)/2);
+}
+template < typename T, T base, std::size_t exponent >
+using pow_ = std::integral_constant<T, pow(base, exponent)>;
```
HannesSommer: I don't see the need for this alias template. Why not use the pow function itself instead?

pointmatcher/DataPointsFilters/utils/utils.h
```diff
@@ -41,6 +41,18 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 namespace utils {
 
+template<class T>
+inline constexpr T pow(const T base, const std::size_t exponent)
+{
+    // (parentheses not required in next line)
```
HannesSommer: This is overcomplicated and the comment superfluous. A good example of (premature optimization](http://wiki.c2.com/?PrematureOptimization). I'd just go for:
```c++
return exponent == 0 ? 1 : base * pow(base, exponent - 1)
```
Consider that every non-empty line of code (including comment lines!) comes at a costs over its entire lifetime as additional load for maintenance and debugging and thus should justify this cost.

utest/ui/DataFilters.cpp
```diff
@@ -445,87 +445,47 @@ TEST_F(DataFilterTest, OctreeGridDataPointsFilter)
...
-		params["buildParallel"] = "1";
-		
-		octreeFilter = 
-				PM::get().DataPointsFilterRegistrar.create("OctreeGridDataPointsFilter", params);
-	
-		const DP filteredCloud = octreeFilter->filter(cloud);
-		
-		//Check number of points
-		EXPECT_GT(cloud.getNbPoints(), filteredCloud.getNbPoints());
+	for(const int meth : {0,1,2})
```
HannesSommer: Yes, that is what I had in mind. I hope you also like it better this way..

pointmatcher/DataPointsFilters/utils/utils.h
```diff
@@ -41,6 +41,18 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 namespace utils {
```
HannesSommer: I'd choose this namespace more specific to Pointmatcher / Octree. What about `PointMatcherSupport` as in https://github.com/ethz-asl/libpointmatcher/blob/master/pointmatcher/PointMatcher.h#L78 ?

pointmatcher/DataPointsFilters/utils/octree.h
```diff
@@ -127,41 +118,44 @@ class Octree
...
+	Octree_();
+	Octree_(const Octree_<T,dim>& o); //Deep-copy
+	Octree_(Octree_<T,dim>&&o);
 	
-	virtual ~Octree();
+	virtual ~Octree_();
 	
-	Octree<T>& operator=(const Octree<T>&o); //Deep-copy
-	Octree<T>& operator=(Octree<T>&& o);
+	Octree_<T,dim>& operator=(const Octree_<T,dim>&o);//Deep-copy
```
HannesSommer: A minor general remark: you are not very consistent with the spaces around operators and the `,`. The pointmatcher convention are if I'm right: single space before and after every binary, ternary operator symbols and single space after each `,`. Unary operators are different but I didn't observe issues. I'm not going to mark that everywhere. Also, sometimes there are lines purely consisting of white spaces. They should just be empty (special case of the "no trailing white spaces"-rule).

pointmatcher/DataPointsFilters/OctreeGrid.cpp
```diff
@@ -171,9 +174,11 @@ bool OctreeGridDataPointsFilter<T>::CentroidSampler::operator()(Octree<T>& oc)
 		//retrieve index from lookup table if sampling in already switched element
 		if(std::size_t(d)<idx)
 			j = mapidx[d];
-			
+		
+		T acc[featDim-1] = {0.};
```
HannesSommer: Jenkins is not happy because of :
```
/home/jenkins/workspace/libpointmatcher/pointmatcher/DataPointsFilters/OctreeGrid.cpp:178:25: error: variable-sized object 'acc' may not be initialized
   T acc[featDim-1] = {0.};
                         ^
```
Btw. variable length arrays such as this one never made it into the c++ standard (AFAIK). It exists as compiler specific extensions and as part of c99. I recommend using `std::vector`. I know, it is sad to accept the additional malloc and free, but compared to the cost of the rest of the function it does not contribute significantly.

pointmatcher/DataPointsFilters/utils/utils.h
MathLabu: I took the code from [here](https://stackoverflow.com/a/16443849).

The author suggest that 
>the compiler isn't required to do constant-initialization at compile-time, but if you use the result of a constexpr function as a non-type template argument, it has to compute the result at compile-time.

I should cite the link in a comment above the function, I forgot it !

pointmatcher/DataPointsFilters/utils/utils.h
MathLabu: Same as above, one of the author suggest using [Exponentiation by squaring](https://en.wikipedia.org/wiki/Exponentiation_by_squaring). But I'm not sure of the benefits, a simpler version should do the same job considering that the evaluation is at the compile time !

HannesSommer: Yes, assuming compile time evaluation it only has a very small effect on the compile time (but for your examples 2, and 3 it might actually be slower with the squaring!) . 

You don't need to care whether this is actually evaluated at compile time or runtime, because it is only about initializing very few global variables (typically 1 or 2 in a given program). 

Again, beware of the big mistake of premature optimization: applied to this case: if you don't have a performance problem with the simple `constexp pow` function, don't go for something more complicated. And trust me, you won't have a performance issue with this one!

pointmatcher/DataPointsFilters/utils/utils.h
MathLabu: No problem for the namespace. Should I keep the functions id the `utils.h` header or move them elsewhere ?

pointmatcher/DataPointsFilters/utils/utils.h
```diff
@@ -39,7 +39,18 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
-namespace utils {
+namespace PointMatcherSupport
+{
+
+template<class T>
+inline constexpr T pow_(const T base, const std::size_t exponent)
+{
+    return (exponent == 0 ? 1 : base * pow_(base, exponent - 1));
+}
+/* use template to force compile time evaluation : https://stackoverflow.com/a/16443849 */
```
HannesSommer: You really don't need that :). 
First, it would be wrong to optimize away the few microseconds on program load (they are no worth the extra complication).  
Second, the way you use pow forces compile time anyway, since you use `nbCells` in context that require compile time evaluation, e.g. here https://github.com/ethz-asl/libpointmatcher/pull/255/files#diff-0e2e58e13d0edca69e786bb76e564ac1R101.

pointmatcher/DataPointsFilters/utils/utils.h
```diff
@@ -39,7 +39,18 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
 #include <algorithm>
 
-namespace utils {
+namespace PointMatcherSupport
+{
+
+template<class T>
+inline constexpr T pow_(const T base, const std::size_t exponent)
+{
+    return (exponent == 0 ? 1 : base * pow_(base, exponent - 1));
```
HannesSommer: Superfluous outer most parenthesis.

pointmatcher/DataPointsFilters/utils/octree.hpp
```diff
@@ -153,457 +151,238 @@ Octree<T>& Octree<T>::operator=(Octree<T>&&o)
...
 		{
-			XYZ{-0.5, -0.5, -0.5},
-			XYZ{+0.5, -0.5, -0.5},
-			XYZ{-0.5, -0.5, +0.5},
-			XYZ{+0.5, -0.5, +0.5},
-			XYZ{-0.5, +0.5, -0.5},
-			XYZ{+0.5, +0.5, -0.5},
-			XYZ{-0.5, +0.5, +0.5},
-			XYZ{+0.5, +0.5, +0.5}
+			{-0.5, -0.5, -0.5},
```
HannesSommer: Ah, much better :).

pointmatcher/DataPointsFilters/utils/octree.hpp
```diff
@@ -153,457 +151,238 @@ Octree<T>& Octree<T>::operator=(Octree<T>&&o)
...
+				this->cells[i] = new Octree_<T,dim>();
+				//Assign depth
+				this->cells[i]->depth = this->depth+1;
+				//Assign parent
+				this->cells[i]->parent = this;
+				//next call is not parallelizable
+				this->cells[i]->build(pts, std::move(sDatas[i]), std::move(boxes[i]), maxDataByNode, maxSizeByNode, false);	
+			};
 		
+		if(not parallelBuild)
```
HannesSommer: Isn't that beautifully readable?

One last minor here: I recommend removing (almost blindly) outer most logic negations from if-conditions while exchanging the branch bodies iff the if statement has an else branch. 
This makes it even more readable (because you remove one operation from the users mind).

pointmatcher/DataPointsFilters/utils/octree.hpp
```diff
@@ -153,457 +151,238 @@ Octree<T>& Octree<T>::operator=(Octree<T>&&o)
...
-	DataContainer datas = TO_DATA(pts, indexes);
+	DataContainer datas = toData(pts, indexes);
 	
 	//build
-	bool ret = true;
-	if(parallel_build) 
-		ret = this->build_par(pts, std::move(datas), std::move(box), maxDataByNode);
-	else 
-		ret = this->build(pts, std::move(datas), std::move(box), maxDataByNode);
+	bool ret = this->build(pts, std::move(datas), std::move(box), maxDataByNode, maxSizeByNode, parallelBuild);
```
HannesSommer: Meaningless variable. You can just inline it: 
```c++
return this-> ...
```

utest/ui/DataFilters.cpp
```diff
@@ -445,87 +445,47 @@ TEST_F(DataFilterTest, OctreeGridDataPointsFilter)
...
+				params["samplingMethod"] = toParam(meth);
+				params["buildParallel"] = "1";
 	
-		//Validate transformation
-		icp.readingDataPointsFilters.clear();
-		addFilter("OctreeGridDataPointsFilter", params);
-		//validate2dTransformation();
-		validate3dTransformation();
-	}
+				octreeFilter =
```
HannesSommer: Why this line break?

pointmatcher/DataPointsFilters/utils/octree.hpp
MathLabu: Sure, I used to put the statement that occurs the most often in the _if_ statement, but indeed it's more readable for the user that way. I'll take that into account for future development!

pomerlef:
Are we alright with this PR?

HannesSommer:
From a code quality point of view it is probably ready (from what I read). I did not look for correctness issues. That is why I did not approve.

pomerlef:
Ok, I saw a couple of graphs with the decomposition. Enough to convince me to go ahead.

PR #256, opened 2018-06-07
Add new filter based on Normal Space sampling

MathLabu:
I added a new filter feature based on Normal Space Sampling (NSS) from _S. Rusinkiewicz and M. Levoy, “Efficient Variants of the ICP Algorithm,” in Proceedings Third International Conference on 3-D Digital Imaging and Modeling, 2001, pp. 145–152_. 

----
The algorithm works as follow:
1. Construct a set of buckets in the normal-space (stocked in a `std::vector`) 
1. Then put all points of the data into buckets based on their normal direction; 
1. Finally, uniformly pick points from all the buckets until the desired number of points is selected.

**Remark:** a point is randomly picked in a bucket that contains multiple points. 
**Remark:** the uniform sampling is based on a standard Mersenne twister engine

---
As the normals are supposed normed, the _n_-space can be represented by polar coordinates, with:
- _theta_, the polar angle in [0 ; pi]
- _phi_, the azimuthal angle in [0 ; 2pi]
- _r_=1, the radius can be omitted

Resources to better understand uniform sampling in normal-space can be found [here](http://corysimon.github.io/articles/uniformdistn-on-sphere/).

---
I had a utest to validate the filter. Some results can be visualized by: 

nss
where the left-white point cloud is the normal distribution of the original point cloud,
where the right-red point cloud is the normal distribution of the sampled point cloud

---
I'm open to any suggestion to improve/clean the filter.

BTW, if someone thinks it's better to `sort` the indexes before reinserting the points at the beginning of the point-cloud than to maintain the old indexes in a `map`, I'm interested to discuss it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -315,7 +315,8 @@ set(POINTMATCHER_SRC
 	pointmatcher/DataPointsFilters/CutAtDescriptorThreshold.cpp
 	pointmatcher/DataPointsFilters/Elipsoids.cpp
 	pointmatcher/DataPointsFilters/Gestalt.cpp
-	pointmatcher/DataPointsFilters/OctreeGrid.cpp		
+	pointmatcher/DataPointsFilters/OctreeGrid.cpp
+	pointmatcher/DataPointsFilters/NormalSpace.cpp	
 )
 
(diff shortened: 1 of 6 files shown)


pomerlef:
Don't forget to add the reference here too.

PR #258, opened 2018-06-20
Add filter based on Covariance Sampling

MathLabu:
I added a new filter feature based on Covariance Sampling (CovS) from _N. Gelfand, L. Ikemoto, S. Rusinkiewicz, and M. Levoy, “Geometrically stable sampling for the ICP algorithm,” in Fourth International Conference on 3-D Digital Imaging and Modeling, 2003. 3DIM 2003. Proceedings., 2003, pp. 260–267._

---
The filter analyses the force (_t-normals_: **n**) and the torque (_r-normals_: **n x p**) to select geometrically stable points that can bind the rotational components as well as the translational. Unlike the original article, we match the point-cloud with itself (considering then an overlap of 100%)

---
Three methods can be used to balance rotation and translation through torque normalization (L):
- L=1 (no normalization): more _t-normals_
- L=Lavg (average distance to centroid) : same contribution for _t-normals_ and _r-normals_
- L=Lmax (in unit ball): more _r-normals_

covs2
with, L=1 in blue, L=Lavg in yellow and L=Lmax in red

---
**Remark:** the filter is not very well suited for large scan with uneven density, it is preferably to use it for computer vision applications, or small scan.


pomerlef:
Could you add your images/comments to the doc too: https://github.com/ethz-asl/libpointmatcher/blob/master/doc/Datafilters.md

Same goes for the other filters.

Issue #262, opened 2018-07-02
first time comiplation error


MahdiSMIDA:
OctreeGrid.cpp:295:42: error: cannot call member function ‘S PointMatcherSupport::Parametrizable::get(const string&) [with S = int; std::__cxx11::string = std::__cxx11::basic_string<char>]’ without object
   const int sm = Parametrizable::get<int>("samplingMethod");

Name            | Version
----------------|-------------------------------
ubuntu:         | Description: Ubuntu 17.10
architecture:   | 64-bit
gcc:            |
git:            | git version 2.14.1
cmake:          | cmake version 3.9.1 CMake suite maintained and supported by Kitware (kitware.com/cmake).
boost:          | Version: 1.62.0.1
eigen3:         | Version: 3.3.4-3
doxygen:        | Version: 1.8.13-9

pomerlef:
@MathLabu Could you have a look? It seems to be from `OctreeGrid.cpp`.

MathLabu:
Hi!

I can't see anything wrong with this [line](https://github.com/ethz-asl/libpointmatcher/blob/master/pointmatcher/DataPointsFilters/OctreeGrid.cpp#L295): as we only call the super method [`get`](https://github.com/ethz-asl/libpointmatcher/blob/master/pointmatcher/Parametrizable.h#L173) of the `Parametrizable` class, you should not have the: 
> error: cannot call member function "..." without object

kind of error, since `OctreeGridDataPointsFilter<T>` inherits of `PointMatcher<T>::DataPointsFilter` which inherits of `Parametrizable` (see [here](https://github.com/ethz-asl/libpointmatcher/blob/master/pointmatcher/PointMatcher.h#L433))... 

As you are on Ubuntu 17.10, I suppose your gcc version is something like 7.x? If there is something wrong with this call, you should have an error [here](https://github.com/ethz-asl/libpointmatcher/blob/master/pointmatcher/DataPointsFilters/MaxPointCount.cpp#L47) too since we call the same method the same way.

Maybe if you provide more information I could find if there is something wrong, but with just this error, I can't see anything wrong, sorry!

tobiasneumann:
I can reproduce this with Fedora 27 and gcc 7.3.1.

I also don't understand how that could work without an instance.
Method `get` calls `getParamValueString` which uses the `parameters` and `parametersUsed` member.

[...]

PR #263, opened 2018-07-04
[fix] call of super method for newer gcc version

tobiasneumann:
Fixes issue #262
thanks to @vmatare for pointing in the right direction


pomerlef:
ok to test

MathLabu:
Should this patch be applied [here](https://github.com/ethz-asl/libpointmatcher/blob/master/pointmatcher/DataPointsFilters/MaxPointCount.cpp#L47) too? Or since the variable has no `const` qualifier it's not a problem?

HannesSommer:
I'd recommend to use either 
1. a more explicit way to access a templated method with a _dependent name_ (which could be considered the underlaying problem here):
```c++
this->template get<int>(...)
```
2. or explicit scoping for the non-static member resolution:
```c++
this->Parametrizable::get<int>(...)
```

and do that for all calls ('agree with @MathLabu ; btw. constness is irrelevant here).

### Explanation of 1.
*  `this->` is the "official" way to mark the name (`get`) as a dependent name (=>the compiler has to wait for instantiation before rejecting the name, which it won't because then it will consider all effective super classes)
* `template` is used to disambiguate the dependent name `get` such that the compiler knows it is a template.

### Explanation of 2.
This is basically the more verbose version of 
```c++
Parametrizable::get<int>(...)
```
stating that the compiler should look for non static members for the this-pointer. Through explicitly specifying the class from which to take the member name, `get`, is non-dependent (in this case) and therefore the disambiguation (with `template`) is not required.
I don't know why the new gcc does not seem to accept the short version (current master) anymore but I'm pretty confident that it'll accept 2.. 
(Does somebody know about an updated language specification here?).

### Why prefer 2. over this PR's suggestion:
```c++
((Parametrizable*)this)->get<int>(...)
```
is technically a cast (a c-style btw., a `static_cast` would be more c++). However, casting is not what this is about. Instead it is about scoping, which IMO it is cleaner to use the scope resolution operator, `::`.

### 1. vs. 2.
Solution 1. is quite different from explicit scoping, 2., (also current solution): With explicitly specifying the class from which to take the member (its name might no longer be dependent - as it is here) and therefore the disambiguation (`template`) is not required, which is quite nice of course.

However, 2.
* somehow obfuscates the underlying problem (dependent name of a template function) since explicit parent scoping typically indicates ambiguities between multiple base classes, for which it is the only solution.
* blocks the (static) inheritance (in contras to 1.), i.e. if the current class (of `this`) started to shadow the `get`-template with its own implementation, then this call would remain unaffected which would probably be unintended.
* is effectively a duplication of a parent class specification, i.e. once the parent class gets changed this call must be adapted.

#### Conclusion
The decision should be based on what is intended here: explicitly take the `get` from `Parametrizable` or somehow from the class hierarchy. I'm not sure here. But if an explicit delegation to the specific class is intended then making it a base class in the first place is confusing (might still be justified, since c++ is a bit nasty here).

@pomerlef, maybe it should be a maintainer decision?  I would suggest 1.  (but it is a tough decision :) )

pomerlef:
Thanks @HannesSommer for the deeper analysis. It seems that there is a preference for solution 2 over here.

@tobiasneumann would it possible to review the PR?

tobiasneumann:
Thank you for the good feedback.
@MathLabu this seems not to be a problem, but I would also change it for consistency reasons.

@HannesSommer and @pomerlef solution 2 dos not compile with the same error as the current master
`‘get<uint8_t>’ is not a member of ‘CovarianceSamplingDataPointsFilter<T>’`
Solution 1 does compile and I agree that this is nicer code than my original PR, I've updated the PR.

HannesSommer:
@tobiasneumann , thanks for trying out. I'm astonished that 2. doesn't work. Just to be sure: you've tried 
```c++
this->Parametrizable::get<int>(...)
```
and then the compiler is complaining that `CovarianceSamplingDataPointsFilter<T>` is missing `get`? 
Is this a compiler bug? Amazing :).
Or was it a misunderstanding and you tried in fact 
```c++
this->CovarianceSamplingDataPointsFilter<T>::get<int>(...)
```
In that case the error message would make sense to me.

tobiasneumann:
@HannesSommer, yes that is what I did. I committed [this](https://github.com/tobiasneumann/libpointmatcher/commit/160c28263fd3fb413c97670c7c288fefdafd1cd0) which is the applied change that isn't working (just to make sure we speak about the same).

With this applied, the compilation error is (copied in greater extent)

```
libpointmatcher/pointmatcher/DataPointsFilters/MaxPointCount.cpp: In constructor ‘MaxPointCountDataPointsFilter<T>::MaxPointCountDataPointsFilter(const Parameters&)’:
libpointmatcher/pointmatcher/DataPointsFilters/MaxPointCount.cpp:47:32: error: ‘get<size_t>’ is not a member of ‘MaxPointCountDataPointsFilter<T>’
   seed = this->Parametrizable::get<size_t>("seed");
```

HannesSommer:
Thanks again. Fascinating! This seems to me like a compiler bug: I could not find any explanation from the standard and clang 6 compiles it just fine.
Probably, we should stop investigating and go for the alternative 1., i.e. your current code in this PR.

tobiasneumann:
@pomerlef I'm sorry, I'm not quite sure what you're asking, could you please elaborate your question?
I tried to make sure that the code and commit msgs are consistent with this project.

pomerlef:
Yes, it was not clear.

I was referring to the comment of @MathLabu about other call to super template  (ex.:
https://github.com/ethz-asl/libpointmatcher/blob/master/pointmatcher/DataPointsFilters/MaxPointCount.cpp#L47)

tobiasneumann:
Yes I included this when I updated the PR according to @HannesSommer suggestions. Even so gcc did not complain about it before. And I've just checked, this seems to be the only occurrence of `Parametrizable::get` in method bodies.

I did not change the calls of `Parametrizable::get` when used in the initializer lists, because this does not cause problems with gcc. If you would like that to be changed as well I can add this to the PR, its only a simple search+replace, but it would increase the PR diff from 3 lines to 150.

@pomerlef which would you prefer?

pomerlef:
That's good for me then!

PR #270, opened 2018-07-30
Distance limit filter introduced to replace both min and max distance filters.

simonpierredeschenes:


pomerlef:
Add to whitelist

pomerlef:
Thanks @simonpierredeschenes !

PR #271, opened 2018-08-01
Add error message when non-rigid transformations are detected.

simonpierredeschenes:
…gonal matrices could not have a determinant equal to -1.


simonpierredeschenes:
Yes it is a reflexion. In my case, it is around the z axis. I stumbled upon this bug when I fixed a 2D lidar upside down and tried to do a map with it. The upside down frame causes a rotational matrix with a determinant equal to -1.

PR #278, opened 2018-09-26
Fix var trimmed

PhiBabin:
The `VarTrimmedDistOutlierFilter` suffered from an implementation error, where the sorted errors was not commutativity sum (see [the original paper](https://arxiv.org/abs/cs/0606098) from more details).


pomerlef:
add to whitelist

Issue #273, opened 2018-08-10
SegFault: Double free when passing same filter to reference and reading


MathLabu:
When passing the same instance of a`PM::DataPointsFilter*` to the reading filters queue and the reference filters queue:
```c++
PM::DataPointsFilter* df = PM::get().DataPointsFilterRegistrar.create(name, params);

referenceDataPointsFilters.clear();
icp.readingDataPointsFilters.clear();

icp.referenceDataPointsFilters.push_back(df);
icp.readingDataPointsFilters.push_back(df);
```
a _segfault_ is encountered due to a double free when cleaning those two queues.

---
Maybe it would be better that the `PM::get().DataPointsFilterRegistrar.create` return a `shared_ptr` instead of a raw pointer, and the `push_back` function could also take a `shared_ptr`. 
We could abstract the underlying type by using an alias `PM::DataPointsFilterPtr` instead of using the raw pointer version.

PR #284, opened 2018-10-08
Replaced raw pointers and boost smart pointers by std raw pointers

simonpierredeschenes:


simonpierredeschenes:
This solves issue #273

PR #289, opened 2018-10-17
[WIP] Add clang-format

PhiBabin:
- [X] add clang-format
- [ ] add git hook

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,64 @@
+---
+BasedOnStyle: Mozilla
+AccessModifierOffset: '-4'
+AlignConsecutiveAssignments: 'false'
+AlignConsecutiveDeclarations: 'false'
+AlignEscapedNewlinesLeft: 'true'
+AlignOperands: false
+AlignTrailingComments: 'false'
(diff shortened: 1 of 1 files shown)


YoshuaNava:
@PhiBabin Can you describe this clang format? Does it cover all the usual conventions from the existing codebase?

stephanemagnenat:
As I was the lead designer of both projects from a software engineering point of view, I indeed used the same conventions for both, so if the the clang-format version of Aseba matches my original conventions, it should work. However, I do not know whether the Aseba project recently diverged convention-wise.

YoshuaNava:
@stephanemagnenat Thank you for your reply. I understand.
I used https://text-compare.com/ to compare the clant-format file to the one from Aseba, and they do match.

I personally use the VSCode clang format plugin to fix formatting of source code. Do you use any specific tool to apply the formatting to the Aseba files?

YoshuaNava:
After reading about this topic a bit I have come to the realization that it might be better to apply formatting per commit / PR, rather than to the whole repo.

pomerlef:
Alright, so we will close it for now.

PR #314, opened 2019-01-23
[feature] Add interface to inform if we reached the maximum number of allowed iterations

Ellon:
Hi,

This is a (very late) followup of #163.

I just came back to use libpointmatcher again and I still need to know if my ICP stopped because it reached the maximum number of allowed iterations or not. Since my previous solution was not accepted, I came about with a simpler one that adds an interface where the user can ask if we reached the limit of iterations. I still use an exception internally to be able to flag the limit being reached, but this is not propagated to the one calling the ICP method. This way we avoid breaking code using libpointmatcher.

I don't think there was any other input about the information users need (neither on #163 nor #23). On my side I just need this info, so I can avoid shipping a modified version of libpointmatcher along with my code. :smile: I would be glad if you could accept it.

Best,

### DIFF
--- a/pointmatcher/ICP.cpp
+++ b/pointmatcher/ICP.cpp
@@ -64,7 +64,8 @@ InvalidModuleType::InvalidModuleType(const std::string& reason):
 template<typename T>
 PointMatcher<T>::ICPChainBase::ICPChainBase():
 	prefilteredReadingPtsCount(0),
-	prefilteredReferencePtsCount(0)
+	prefilteredReferencePtsCount(0),
+	maxNumIterationsReached(false)
 {}
 
 //! virtual desctructor
@@ -177,6 +178,13 @@ template<typename T>
 unsigned PointMatcher<T>::ICPChainBase::getPrefilteredReferencePtsCount() const
 {
 	return prefilteredReferencePtsCount;
+}
+
+//! Return the flag that informs if we reached the maximum number of iterations during the last iterative process
+template<typename T>
+bool PointMatcher<T>::ICPChainBase::getMaxNumIterationsReached() const
+{
+	return maxNumIterationsReached;
 }
 
 //! Instantiate modules if their names are in the YAML file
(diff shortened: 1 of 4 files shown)


PhiBabin:
I think it would be preferable, if the `CounterTransformationChecker` had a boolean attribute `max_iteration_reached`. Then you can access this attribute by iterating on the different transformation checkers in  `icp.transformationCheckers` and finding the `CounterTransformationChecker` instance. With this instance you can access the `max_iteration_reached` attribute.
This solution is terrible from a api point of view and it's not really user friendly. But it prevents a dependencies between the main ICP module and a particular filter/transformationChecker.

pomerlef:
ok to test

Ellon:
@PhiBabin I agree with you on both aspects of your proposed solution: not as user friendly but it prevents dependencies.

I implemented your idea in a local test branch, but I'm having problem getting a pointer to the `CounterTransformationChecker` instance. AFAIK, this class is defined on [TransformationCheckersImpl.h](https://github.com/ethz-asl/libpointmatcher/blob/master/pointmatcher/TransformationCheckersImpl.h#L58) and this header is not installed along with libpointmatcher files. So I'm not able to do something like this:

```c++
#include <pointmatcher/TransformationCheckersImpl.h>
(...)
using CounterTransformationChecker = typename TransformationCheckersImpl<T>::CounterTransformationChecker;
for (const auto checker_ptr : icp_.transformationCheckers) {
  auto counter_checker_ptr = std::dynamic_pointer_cast<CounterTransformationChecker>(checker_ptr);
  if (counter_checker_ptr and counter_tr_checker_ptr->isMaxIterationReached()) {
    // Do something...
  }
}
```

Besides the associated source file, the other places `TransformationCheckersImpl.h` is included are [ICP.cpp](https://github.com/ethz-asl/libpointmatcher/blob/master/pointmatcher/ICP.cpp#L46) and [Registry.cpp](https://github.com/ethz-asl/libpointmatcher/blob/master/pointmatcher/Registry.cpp#L44) and both are internal sources to the library.

How could a user perform this dynamic cast without access to the class definition?

PhiBabin:
@Ellon You are completely right, I had forgotten that libpointmatcher doesn't install the header of the different modules. In that case, I think that your PR is a way better solution to the problem of detecting the max number of iteration. `CounterTransformationChecker` is used by almost all yaml files, so I don't think the added dependencies to the main ICP will be a big issue.

PR #320, opened 2019-02-12
Added documentation for people using ROS.

simonpierredeschenes:


pomerlef:
alright, should we mention if the install it through `apt-get` or from sources?

simonpierredeschenes:
I addressed your comment in the last commit

PR #327, opened 2019-04-02
Fix/normal space hashing

jsburklund:
Fix hashing issues.

Explicity use the correct rounding direction so that enough buckets are allocated, and handle wrapping theta appropriately at Pi. Previously, if theta=Pi, the hash returned an index out of bounds of the preallocated memory and caused a segfault.


simonpierredeschenes:

We'll look at this PR more in details at the end of this week.

simonpierredeschenes:
Looks good to me, to merge.

PR #328, opened 2019-04-02
Improve speed of Normal Space filtering

jsburklund:
Improves the speed of Normal Space binning and selection.

Use a more efficient sampling method instead of rejection sampling. Improves runtime speed and determinism by not attempting to sample low probability events. Remove empty bins and randomly select from only the remaining available bins. Points are added into each bin initially in a random order, and are now popped off the back of a list of points inside the bucket.


simonpierredeschenes:

We'll look at this PR more in details at the end of this week.

simonpierredeschenes:
@jsburklund Wow, your implementation is really faster than the old one! It will be merged soon. Thanks again for your contribution!

PR #367, opened 2019-12-16
Updated - Point-to-plane Covariance Code

goldbattle:
In the process of trying to get an accurate covariance from this library I have updated the point-to-plane minimizer to produce it. I was unable to figure out how the original covariance code was derived and the second derivative in respect to the measurement and then state did not make sense. If someone could point me in the direction to get the same equations as the original repo.

This is based off of this work and repository which does symbolic differentiation:
> Prakhya, Sai Manoj, et al. "A closed-form estimate of 3D ICP covariance."
> 2015 14th IAPR International Conference on Machine Vision Applications (MVA). IEEE, 2015.
> http://www.mva-org.jp/Proceedings/2015USB/papers/14-27.pdf
> https://github.com/saimanoj18/3d-icp-covariance

I re-ran the scripts and also changed it to get a 2d icp covariance using the same methodology.

### DIFF
--- a/doc/matlab/point_to_plane_first_term_2d.m
+++ b/doc/matlab/point_to_plane_first_term_2d.m
@@ -1,0 +1,94 @@
+clc;
+clear; 
+% declare the symbolic variables
+x = sym('x','real');
+y = sym('y','real');
+
+a = sym('a','real');
+
+
+T = [x; y;];
+R = [cos(a) -sin(a);
+     sin(a) cos(a)];
+
+pix = sym('pix','real');
+piy = sym('piy','real');
+
+qix = sym('qix','real');
+qiy = sym('qiy','real');
+ 
+Pi = [pix;piy];
+Qi = [qix;qiy];
+
+nix = sym('nix','real');
+niy = sym('niy','real');
+
+Ni = [nix;niy;];
+
+G = dot((R * Pi + T - Qi), Ni);    %Ni should correspond to Qi
+dG_dx = jacobian(G,x);
+dJ_dx = 2 * dG_dx' * G;
+
+dG_dy = jacobian(G,y);
+dJ_dy = 2 * dG_dy' * G;
+
+
+%% 1,2
+d2J_dx2 = jacobian(dJ_dx,x)
+d2J_dy2 = jacobian(dJ_dy,y)
+%% 4,5,6 , 34,35,36
+d2J_dydx = jacobian(dJ_dx,y)
+d2J_dxdy = jacobian(dJ_dy,x)
+
+
+
+
+%% 7
+dG_da = jacobian(G,a);
+dJ_da = 2 * dG_da' * G;
+d2J_da2 = jacobian(dJ_da,a)
+
+
+%% (10,11) , (12,13) , (14,15)
+d2J_dxda = jacobian(dJ_da,x)
+d2J_dadx = jacobian(dJ_dx,a)
+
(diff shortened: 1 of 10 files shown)


pomerlef:
ok to test

YoshuaNava:
Hi,
This looks like a really nice contribution. Thank you!

I wanted to point out, nonetheless, that I recently read [this paper](https://arxiv.org/abs/1909.05722) where different methods to estimate ICP Covariance are compared (Censi, CELLO-3D and their own). Their conclusion is that Censi's approach tends to be over-confident, as it ignores the dependence of the ICP final estimate distribution on the initial guess provided, which is a key factor.

The author of the above paper is @mbrossar. He has provided an example implementation here: https://github.com/CAOR-MINES-ParisTech/3d-icp-cov

YoshuaNava:
@goldbattle Do you have any metrics about:
* Usage of computational resources (CPU/memory) of this new covariance estimation method?
* Quality of the estimates.

And the application of the changes you made here to a project?

This is quite an interesting contribution.

YoshuaNava:
> @goldbattle Do you have any metrics about:
> 
> * Usage of computational resources (CPU/memory) of this new covariance estimation method?
> * Quality of the estimates.
> 
> And an example application of the changes you made to a project? This is quite an interesting contribution and it would be great to look at it closely.

Ping @goldbattle :)

goldbattle:
I have not performed a very thorough evaluation as in [Martin's paper](https://ieeexplore.ieee.org/abstract/document/8954799). Censi’s covariance seems to give reasonable results such that I don't need to inflate/deflat it in order to incorporating it into a probabilistic SLAM framework. I have not verified as in Martin's work, so probably in the edge cases it breaks down and is very poor.

As in terms of time, for 2d ICP it takes about 3-4ms single threaded, set using the `OMP_NUM_THREADS=1` environmental variable, for alignment of two clouds with 500 points. I have a Intel(R) Xeon(R) CPU E3-1505M v6 @ 3.00GHz. I don't have any comparison to non-covariance based since that isn't a usable result for me.

YoshuaNava:
@goldbattle I will try your method on 3D point clouds.

Regarding the code, would it be possible to maybe implement a switch? So that we have different methods to estimate the covariance?

goldbattle:
I am not even sure how the old results were derived. If someone can point me to the derivation then that would be ideal, as I didn't get anything close to what was implemented for the small parts that I did original derive by hand.

pomerlef:
I'm 100 % sure, but it might be from here : https://arxiv.org/pdf/1410.7632.pdf
but the math are from here: https://www.researchgate.net/publication/276269511_A_Closed-form_Estimate_of_3D_ICP_Covariance

But, that's on top of my head. I might be wrong.

YoshuaNava:
Update: I have not forgotten about this PR, but I'm currently very busy. I'm in contact with @goldbattle to introduce a way of testing this PR and others that might change the way covariances are computed.

I estimate that I work on this again in April.

pomerlef:
It's been a while, where are we with this?

YoshuaNava:
@pomerlef Sorry for the delay. I'll look into this very soon.

pomerlef:
Unfortunately, validating a covariance is still ill defined (at least for my personal point of view). In these cases, it might be better to check what is current output and making sure that it's not influenced in the future by an external modification. That's how the current filters are validated in terms of impact on ICP.

PR #374, opened 2020-04-22
Feature/geometry data points filter for master

kubelvla:
This filter reproduces our heuristic intended for estimating radio signal attenuation through point clouds (GPS and UGF signals).


pomerlef:
ok to test

kubelvla:
Jenkins has some error related to be unable to read some output log file (java exception in the end of the console output). Seems to me more like Jenkins configuration problem than the actual build and test problem...

pomerlef:
that's a recurring problem and why we want to transfer ownership

YoshuaNava:
I'm a user of this package and would like to help reviewing the changes. I will proceed to that.

pointmatcher/DataPointsFilters/Geometry.cpp
```diff
@@ -0,0 +1,170 @@
...
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+*/
+#include "Geometry.h"
+
+// Eigenvalues
+#include "Eigen/QR"
+#include "Eigen/Eigenvalues"
```
YoshuaNava, line 39: Does Libpointmatcher have any policy regarding internal/external includes?

For example, https://github.com/ethz-asl/segmap/wiki/Contributing-to-SegMap represents external include paths between <>, and internal ones with ""

pointmatcher/DataPointsFilters/Geometry.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+#include "PointMatcherPrivate.h"
+#include "IO.h"
+#include "MatchersImpl.h"
```
YoshuaNava, line 43: The order of these includes is not alphabetical. In the near future we could have a linter with a rule to fix this automatically.

pointmatcher/DataPointsFilters/Geometry.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+#include <boost/format.hpp>
```
YoshuaNava, line 45: The format of this include is different to the previous ones. Any particular reason?

pointmatcher/DataPointsFilters/Geometry.cpp
```diff
@@ -0,0 +1,170 @@
...
+void GeometryDataPointsFilter<T>::inPlaceFilter(
+	DataPoints& cloud)
+{
+	typedef typename DataPoints::View View;
+	typedef typename DataPoints::Label Label;
+	typedef typename DataPoints::Labels Labels;
+
+	const int pointsCount(cloud.features.cols());
+	const int descDim(cloud.descriptors.rows());
+	const unsigned int labelDim(cloud.descriptorLabels.size());
```
YoshuaNava, line 83: This line is different to the others.

pointmatcher/DataPointsFilters/Geometry.cpp
```diff
@@ -0,0 +1,170 @@
...
+template<typename T>
+void GeometryDataPointsFilter<T>::inPlaceFilter(
+	DataPoints& cloud)
+{
+	typedef typename DataPoints::View View;
+	typedef typename DataPoints::Label Label;
+	typedef typename DataPoints::Labels Labels;
+
+	const int pointsCount(cloud.features.cols());
+	const int descDim(cloud.descriptors.rows());
```
YoshuaNava, line 82: Why use signed int instead of an unsigned int? You leave half of the bits unused.

I would suggest using size_t: https://www.geeksforgeeks.org/size_t-data-type-c-language/

pointmatcher/DataPointsFilters/Geometry.cpp
```diff
@@ -0,0 +1,170 @@
...
+	const unsigned int labelDim(cloud.descriptorLabels.size());
+
+	// Check that the required eigenValue descriptor exists in the pointcloud
+    if (!cloud.descriptorExists("eigValues"))
+    {
+        throw InvalidField("GeometryDataPointsFilter: Error, no eigValues found in descriptors.");
+    }
+
+    // Validate descriptors and labels
```
YoshuaNava, line 91: Indentation of this line differs.

pointmatcher/DataPointsFilters/Geometry.cpp
```diff
@@ -0,0 +1,170 @@
...
+{
+	typedef typename DataPoints::View View;
+	typedef typename DataPoints::Label Label;
+	typedef typename DataPoints::Labels Labels;
+
+	const int pointsCount(cloud.features.cols());
+	const int descDim(cloud.descriptors.rows());
+	const unsigned int labelDim(cloud.descriptorLabels.size());
+
+	// Check that the required eigenValue descriptor exists in the pointcloud
```
YoshuaNava, line 85: Indentation of this line differs with respect to the code below.

pointmatcher/DataPointsFilters/Geometry.cpp
```diff
@@ -0,0 +1,170 @@
...
+	int insertDim(0);
+	for(unsigned int i = 0; i < labelDim ; ++i)
+		insertDim += cloud.descriptorLabels[i].span;
+	if (insertDim != descDim)
+		throw InvalidField("SurfaceNormalDataPointsFilter: Error, descriptor labels do not match descriptor data");
+
+	// Reserve memory for new descriptors
+	const int dimSphericality(1);
+	const int dimUnstructureness(1);
+	const int dimStructureness(1);
```
YoshuaNava, line 101: The above lines fulfill the same purpose, and even though it is illustrative to write the code like this, more lines of code result in more maintenance.

Would it be possible to replace these lines with a single `unidimensionalDescriptorDimension` const int?

pointmatcher/DataPointsFilters/Geometry.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+	Labels cloudLabels;
+	cloudLabels.push_back(Label("sphericality", dimSphericality));
+	if (keepUnstructureness)
+		cloudLabels.push_back(Label("unstructureness", dimUnstructureness));
+	if (keepStructureness)
+		cloudLabels.push_back(Label("structureness", dimStructureness));
+
+    // Reserve memory
+	cloud.allocateDescriptors(cloudLabels);
```
YoshuaNava, line 115: Different indentation.

pointmatcher/DataPointsFilters/Geometry.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+	// Get the views
+    const View eigValues = cloud.getDescriptorViewByName("eigValues");
+    if (eigValues.rows() != 3)  // And check the dimensions
+    {
+        throw InvalidField("GeometryDataPointsFilter: Error, the number of eigValues is not 3.");
+    }
```
YoshuaNava, line 122: Ditto indentation.

pointmatcher/DataPointsFilters/Geometry.cpp
```diff
@@ -0,0 +1,170 @@
...
+		cloudLabels.push_back(Label("unstructureness", dimUnstructureness));
+	if (keepStructureness)
+		cloudLabels.push_back(Label("structureness", dimStructureness));
+
+    // Reserve memory
+	cloud.allocateDescriptors(cloudLabels);
+
+	// Get the views
+    const View eigValues = cloud.getDescriptorViewByName("eigValues");
+    if (eigValues.rows() != 3)  // And check the dimensions
```
YoshuaNava, line 119: From these lines I can infer that this filter works only in 3D, right?

Maybe the name of the filter class should reflect that. Or at least have a parameter to specify the mode?

pointmatcher/DataPointsFilters/Geometry.cpp
```diff
@@ -0,0 +1,170 @@
...
+    {
+        throw InvalidField("GeometryDataPointsFilter: Error, the number of eigValues is not 3.");
+    }
+
+	sphericality = cloud.getDescriptorViewByName("sphericality");
+	if (keepUnstructureness)
+		unstructureness = cloud.getDescriptorViewByName("unstructureness");
```
YoshuaNava, line 126: Indentation too high.

pointmatcher/DataPointsFilters/Geometry.cpp
```diff
@@ -0,0 +1,170 @@
...
+	if (keepStructureness)
+		structureness = cloud.getDescriptorViewByName("structureness");
+
+    // Iterate through the point cloud and evaluate the geometry
+	for (int i = 0; i < pointsCount; ++i)
```
YoshuaNava, line 131: As before I would suggest using an unsigned int here.

pointmatcher/DataPointsFilters/Geometry.cpp
```diff
@@ -0,0 +1,170 @@
...
+	{
+	    // look at the three eigen values
+        Vector eig_vals_col = eigValues.col(i);
```
YoshuaNava, line 134: Here you don't seem to be looking at the eigenvalues per se, rather extracting a set of eigenvalues from a larger matrix.

pointmatcher/DataPointsFilters/Geometry.cpp
```diff
@@ -0,0 +1,170 @@
...
+        // might be already sorted but sort anyway
+        std::sort(eig_vals_col.data(),eig_vals_col.data()+eig_vals_col.size());
```
YoshuaNava, line 136: Can you apply this operation before? Or prevent running it here?

By keeping it inside you are increasing the complexity of this loop from O(n) to O(n*3)

pointmatcher/DataPointsFilters/Geometry.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+        // finally, evaluate the geometry
+        T sphericality_val;
```
YoshuaNava, line 139: These variables don't follow `camelCasing` for variables, as done in other lines of this file and libpointmatcher source.

pointmatcher/DataPointsFilters/Geometry.cpp
```diff
@@ -0,0 +1,170 @@
...
+        T unstructureness_val;
+        T structureness_val;
+
+        //TODO: Is there a more suitable limit for considering the values almost-zero? (VK)
+        if (fabs(eig_vals_col(2)) < std::numeric_limits<T>::min() or
```
YoshuaNava, line 144: You seem to be evaluating the smallest eigenvalues to determine the span of the points in space?

Can you write a comment explaining what you are doing here? If other developers want to extend your code, that will help them to get up to speed quickly.

pointmatcher/DataPointsFilters/Geometry.cpp
```diff
@@ -0,0 +1,170 @@
...
+        {
+            sphericality_val = std::numeric_limits<T>::quiet_NaN();
+            unstructureness_val = std::numeric_limits<T>::quiet_NaN();
+            structureness_val = std::numeric_limits<T>::quiet_NaN();
+        } else {
+            unstructureness_val = eig_vals_col(0) / eig_vals_col(2);
+            structureness_val =  (eig_vals_col(1) / eig_vals_col(2)) *
+                    ((eig_vals_col(1) - eig_vals_col(0)) / sqrt(eig_vals_col(0)*eig_vals_col(0) + eig_vals_col(1)*eig_vals_col(1)));
+            sphericality_val = unstructureness_val - structureness_val;
+        }
```
YoshuaNava, line 155: What is being done in this block?

pointmatcher/DataPointsFilters/Geometry.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+        // store in the pointcloud
+        (*sphericality)(0,i) = sphericality_val;
```
YoshuaNava, line 158: Is there a way to prevent this kind of access?

pointmatcher/DataPointsFilters/Geometry.cpp
```diff
@@ -0,0 +1,170 @@
...
+		throw InvalidField("SurfaceNormalDataPointsFilter: Error, descriptor labels do not match descriptor data");
+
+	// Reserve memory for new descriptors
+	const int dimSphericality(1);
+	const int dimUnstructureness(1);
+	const int dimStructureness(1);
+
+	boost::optional<View> sphericality;
+	boost::optional<View> unstructureness;
+	boost::optional<View> structureness;
```
YoshuaNava, line 105: What is the advantage of these optionals? How are they mapped in memory?

Considering that libpointmatcher uses very flat representations (Eigen matrices), and speed is important, I would consider using a fixed size array. I assume that would make the resource usage of this filter more predictable.

pointmatcher/DataPointsFilters/Geometry.h
```diff
@@ -0,0 +1,80 @@
+// kate: replace-tabs off; indent-width 4; indent-mode normal
+// vim: ts=4:sw=4:noexpandtab
```
YoshuaNava, line 2: Same comment as the one in the source file: are these lines intended to be there?

pointmatcher/DataPointsFilters/Geometry.h
```diff
@@ -0,0 +1,80 @@
...
+/*
+
+Copyright (c) 2010--2018,
+François Pomerleau and Stephane Magnenat, ASL, ETHZ, Switzerland
+You can contact the authors at <f dot pomerleau at gmail dot com> and
+<stephane at magnenat dot net>
+
```
YoshuaNava, line 9: I would suggest asking the author if you can also add yourself here, not only to ack your contribution, but as a way of informing people of who wrote the file, and maybe providing contact information and a mention of the papers where this method is described.

pointmatcher/DataPointsFilters/Geometry.h
```diff
@@ -0,0 +1,80 @@
...
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+*/
+#pragma once
+
+#include "PointMatcher.h"
+
+#include <vector>
```
YoshuaNava, line 39: This isn't used in the header.

pointmatcher/DataPointsFilters/Geometry.h
```diff
@@ -0,0 +1,80 @@
...
+	typedef Parametrizable::InvalidParameter InvalidParameter;
+	
+	typedef typename PointMatcher<T>::Vector Vector;
+	typedef typename PointMatcher<T>::Matrix Matrix;	
+	typedef typename PointMatcher<T>::DataPoints DataPoints;
+	typedef typename PointMatcher<T>::DataPoints::InvalidField InvalidField;
+
+	inline static const std::string description()
+	{
+		return "This filter computes the level of ‘unstructureness’, ‘structureness’ and ’sphericality’ for each point based on the required eigen values.\n\n"
```
YoshuaNava, line 59: Can you introduce a more in depth explanation?

Judging by the name of the class, I would assume that the filter tells you something about the geometry of the point cloud, but it seems to actually compute descriptors that illustrate the geometrical distribution of points. 

---
Another change I would suggest is:

eigen values -> eigenvalues

pointmatcher/DataPointsFilters/Geometry.h
```diff
@@ -0,0 +1,80 @@
...
+			{"keepUnstructureness", "whether the value of the unstructureness should be added to the pointcloud", "0"},
+			{"keepStructureness", "whether the value of the structureness should be added to the pointcloud", "0"},
+			};
+	}
+	
+	const bool keepUnstructureness;
+	const bool keepStructureness;
+
+    GeometryDataPointsFilter(const Parameters& params = Parameters());
+	virtual ~GeometryDataPointsFilter() {};
```
YoshuaNava, line 77: Instead of an empty destructor can use the `default` keyword?

https://stackoverflow.com/questions/6502828/what-does-default-mean-after-a-class-function-declaration

pointmatcher/DataPointsFiltersImpl.h
```diff
@@ -92,6 +93,11 @@ struct DataPointsFiltersImpl
 	typedef ::CovarianceSamplingDataPointsFilter<T> CovarianceSamplingDataPointsFilter;
 	typedef ::DistanceLimitDataPointsFilter<T> DistanceLimitDataPointsFilter;
 	typedef ::RemoveSensorBiasDataPointsFilter<T> RemoveSensorBiasDataPointsFilter;
+    typedef ::GeometryDataPointsFilter<T> GeometryDataPointsFilter;
```
YoshuaNava, line 96: Indentation.

pomerlef:
ok to test

pointmatcher/DataPointsFilters/Geometry.cpp
```diff
@@ -0,0 +1,170 @@
+// kate: replace-tabs off; indent-width 4; indent-mode normal
+// vim: ts=4:sw=4:noexpandtab
```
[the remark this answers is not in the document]
kubelvla: Hi @YoshuaNava,
thanks a lot for the review! I based the source files on another filter, which had this type of header. Now I've double-checked all the other files and they are all like this. I guess it is a practical thing to have it there. I'll keep it.

pointmatcher/DataPointsFilters/Geometry.cpp
kubelvla: I actually forgot to remove these includes, they are from the original file I took as a model and they are probably not needed at all... Nevertheless, I haven't heard about such internal convention. @pomerlef do you have a strong opinion on this?

pomerlef: no policy was close to most of our policies, so no.

If it help readability, it can be added.

pointmatcher/DataPointsFilters/Geometry.cpp
kubelvla: This is little embarrassing, but none of the includes is actually needed. Deleting them. But in my future stuff, I'll try to keep the groups of includes alphabetically sorted.

pointmatcher/DataPointsFilters/Geometry.cpp
kubelvla: Deleting those as well.

pointmatcher/DataPointsFilters/Geometry.cpp
kubelvla: Good point, they can all be unsigned ints.

pointmatcher/DataPointsFilters/Geometry.cpp
kubelvla: Yep, good point.

pointmatcher/DataPointsFilters/Geometry.cpp
kubelvla: Ok, this will be mess, I'm using Clion and it by default does tab by four spaces...

kubelvla: Ok, not that bad

pointmatcher/DataPointsFilters/Geometry.cpp
kubelvla: yep

pointmatcher/DataPointsFilters/Geometry.cpp
kubelvla: Good point, done.

pointmatcher/DataPointsFilters/Geometry.cpp
kubelvla: Done

pointmatcher/DataPointsFilters/Geometry.cpp
kubelvla: Done

pointmatcher/DataPointsFilters/Geometry.cpp
kubelvla: I think I'll rename the filter to SphericalLevelDataPointsFilter, which kinda indicates 3D.

pointmatcher/DataPointsFilters/Geometry.cpp
kubelvla: Switched everything to tabs intead of spaces...

pointmatcher/DataPointsFilters/Geometry.cpp
kubelvla: yep

pointmatcher/DataPointsFilters/Geometry.cpp
kubelvla: rephrased

YoshuaNava:
@kubelvla Would it be possible to add the unit tests for this new filter?

Additionally, will you add it to the documentation within the scope of this PR?

pointmatcher/DataPointsFilters/Geometry.cpp
kubelvla: This is an important point and not being a computer scientist, I'll use your expertise here @YoshuaNava. The eig_vals_col contains exactly three values, and they might not be sorted (the filter that produces the eigenvalues can sort them, but does not store that information in the point cloud, so I cannot be sure. Something tells me that actually checking whether they are sorted (two operations here in C++ but actually how many in the assembly) will be in similar number of operations as sorting them right away. I think putting the sorting out of the loop would not help, I still need to sort them by the triplets (cols of the matrix). What do you suppose? Thx!

pointmatcher/DataPointsFilters/Geometry.cpp
kubelvla: Fixed

pointmatcher/DataPointsFilters/Geometry.cpp
kubelvla: I've put there a reference to a paper

pointmatcher/DataPointsFilters/Geometry.cpp
YoshuaNava: Does libpointmatcher use tabs in other files?

I would recommend using spaces if possible, because tab width changes depending on your environment.

Ref: https://softwareengineering.stackexchange.com/questions/57/tabs-versus-spaces-what-is-the-proper-indentation-character-for-everything-in-e

pointmatcher/DataPointsFilters/Geometry.cpp
kubelvla: Actually here will be the reference to the paper, the previous block just check for division by zero...

pointmatcher/DataPointsFilters/Geometry.cpp
kubelvla: I've taken this scheme from another filter from the libpointmatcher (sampling surface normal). I understand the "optional" construct as a slightly smarter pointer. It points to a view on the data in the pointcloud, whose structure I don't want to change. Given that in the end all of these are pointers, I don't think there is any significant overhead (the expensive thing is done once in the `cloud.allocateDescriptors(cloudLabels);` call).

pointmatcher/DataPointsFilters/Geometry.h
kubelvla: They are everywhere in the library :D

pointmatcher/DataPointsFilters/Geometry.h
kubelvla: I was actually about to put myself there (in one commit, I had there TODO), but then all the files, even written by our students, have this copyright header. So I was not sure if by changing it, I would not have messed some legal stuff (there is the 2010--2018, so if I put myself there, do I have to somehow extend the copyright or not...). @pomerlef How to add note about the author? Below the copyright text?

pointmatcher/DataPointsFilters/Geometry.h
kubelvla: Removing that, thanks.

pointmatcher/DataPointsFilters/Geometry.h
kubelvla: I think I'll rename the filter to Sphericality, it also suggests the required 3D

pointmatcher/DataPointsFilters/Geometry.cpp
kubelvla: I am afraid that there will always be some kind of dereferencing, since we are working with a view on the point cloud data... The star operator is overloaded from the boost optional thing to make it act as if dereferencing a pointer. I have changed the code to a method .get() call on the object, that look a little better (but still generates warnings for the next two optionals that they might not be initialized).

pointmatcher/DataPointsFilters/Geometry.h
kubelvla: I think that both ways are clear to understand and the empty-body style is common to this library.

pointmatcher/DataPointsFiltersImpl.h
kubelvla: Yep

pointmatcher/DataPointsFilters/Geometry.h
kubelvla: Also added more details on what that actually does.

kubelvla:
For the unit tests, I'll think about what should actually be tested and then ask one of our students to help me. The bibliography will be updated as soon as the paper is out, now waiting for review.

YoshuaNava:
Hi @kubelvla,
Thank you for addressing the comments.

I understand. I would propose to merge as soon as possible if you and the author approves. Hence, can we open an issue about creating tests and documentation for this filter?

pointmatcher/DataPointsFilters/Sphericality.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+/*
+ The SphericalityDataPointFilter estimates a heuristic value which indicates how much local geometry around
+ a given point resemble a plane or a sphere (uniform distribution). The value can lie in <-1,1> interval, -1 for
+ perfect plane, 1 for perfectly uniform distribution. The estimation is based on three eigenvalues coming from another
+ data points filter (this filter can operate only on 3D data). In the case the largest eigenvalue is zero, the filter
+ outputs NaNs. If the middle eigenvalue is zero and the largest one is non-zero, the filter outputs zeros.
+
+ Implemented by Vladimir Kubelka (kubelvla@gmail.com), NORLAB, Universite Laval, 2020
+*/
```
YoshuaNava, line 46: Cool! :)

kubelvla:
@YoshuaNava I'll open the issue with the biblio and unit test for myself.

pointmatcher/DataPointsFilters/Geometry.cpp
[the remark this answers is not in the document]
YoshuaNava: Ok

pointmatcher/DataPointsFilters/Geometry.h
YoshuaNava: Cool! Thank you.

pointmatcher/DataPointsFilters/Geometry.h
YoshuaNava: Ok.

YoshuaNava:
@kubelvla Looks good to me.

YoshuaNava:
@kubelvla What is the status of this PR? Do you have further implementations planned?

kubelvla:
@YoshuaNava Under the condition I made no mistake when implementing the filter, there should be no changes to this filter.

YoshuaNava:
@kubelvla Is this PR ready for merging?

pomerlef:
If there is no objection, I'll merge it tomorrow.

YoshuaNava:
@pomerlef No objection from my side.

PR #378, opened 2020-04-30
Feature/4 dof for point to plane minimizer

kubelvla:
This PR adds the possibility to optimize only translation and yaw angle instead of the full 6DOF pose. A unit test has been also added. Since this feature is a part of research in progress, there are no publications yet. The code has been successfully used during the DARPA SubT challenge, so I consider it quite well tested.


pomerlef:
add to whitelist

examples/data/icp_data/force4DOFForPointToPlaneMinimizer.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+referenceDataPointsFilters:
+  - SurfaceNormalDataPointsFilter:
+      knn: 15
+
+  - ObservationDirectionDataPointsFilter
+
+  - OrientNormalsDataPointsFilter:
+      towardCenter: 1
+
```
YoshuaNava, line 16: Are the extra lines between the reference data points filters intended?

YoshuaNava:
I just approved this PR, it's a really cool contribution.

Just out of curiosity: did you observe any improvement in the CPU usage or accuracy of ICP with respect to full DOF minimization?

examples/data/icp_data/force4DOFForPointToPlaneMinimizer.yaml
kubelvla: I like it better, easier to distinguish between the filters.

kubelvla:
There should not be a significant speedup, the difference is solving a 4x4 matrix instead of 6x6 one. I don't expect much speedup, the heavy lifting is done elsewhere in ICP...

YoshuaNava:
Ok. All good.

YoshuaNava:
@pomerlef Can this PR be merged?

pomerlef:
All good, go ahead

kubelvla:
@simonpierredeschenes Simon Pierre, could you press the button please? Thx

PR #395, opened 2020-08-04
Replaced the remaining raw pointers with shared pointers in Registrar.h

AlexandreG87:
Also replaced the boost.auto macro with the C++11 auto.

### DIFF
--- a/pointmatcher/Registrar.h
+++ b/pointmatcher/Registrar.h
@@ -155,7 +155,7 @@ namespace PointMatcherSupport
 		//! Register a class by storing an instance of a descriptor helper class
 		void reg(const std::string &name, std::shared_ptr<ClassDescriptor> descriptor)
 		{
-			classes[name] = descriptor;
+			classes.insert(std::make_pair(name, descriptor));
 		}
 
 		//! Return a descriptor following a name, throw an exception if name is invalid
--- a/pointmatcher/Registrar.h
+++ b/pointmatcher/Registrar.h
@@ -70,14 +70,14 @@ namespace PointMatcherSupport
 	{
 		InvalidElement(const std::string& reason);
 	};
-	
+
 	//! A factor for subclasses of Interface
 	template<typename Interface>
 	struct Registrar
 	{
 	public:
 		typedef Interface TargetType; //!< alias to recover the template parameter
-		
+
 		//! The interface for class descriptors
 		struct ClassDescriptor
 		{
@@ -90,7 +90,7 @@ namespace PointMatcherSupport
 			//! Return the available parameters for this class
 			virtual const Parametrizable::ParametersDoc availableParameters() const = 0;
 		};
-		
+
 		//! A descriptor for a class C that provides parameters
 		template<typename C>
 		struct GenericClassDescriptor: public ClassDescriptor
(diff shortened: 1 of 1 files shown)


pointmatcher/Registrar.h
```diff
@@ -90,80 +90,79 @@ namespace PointMatcherSupport
...
 			virtual const Parametrizable::ParametersDoc availableParameters() const
 			{
 				return Parametrizable::ParametersDoc();
 			}
 		};
-		
+
 	protected:
-		typedef std::map<std::string, ClassDescriptor*> DescriptorMap; //!< descriptors for sub-classes of Interface, indexed by their names
+		typedef std::map<std::string, std::shared_ptr<ClassDescriptor>> DescriptorMap; //!< descriptors for sub-classes of Interface, indexed by their names
```
YoshuaNava, line 151: Does it make sense to use `using` here?

https://www.internalpointers.com/post/differences-between-using-and-typedef-modern-c

pointmatcher/Registrar.h
```diff
@@ -90,80 +90,79 @@ namespace PointMatcherSupport
...
 	public:
-		//! Destructor, remove all classes descriptors 
-		~Registrar()
-		{
-			for (BOOST_AUTO(it, classes.begin()); it != classes.end(); ++it)
-				delete it->second;
-		}
 		//! Register a class by storing an instance of a descriptor helper class
-		void reg(const std::string &name, ClassDescriptor* descriptor)
+		void reg(const std::string &name, std::shared_ptr<ClassDescriptor> descriptor)
```
YoshuaNava, line 156: Is the descriptor intended to change? Otherwise we could pass a const reference to it.

https://stackoverflow.com/questions/8385457/should-i-pass-a-shared-ptr-by-reference

pointmatcher/Registrar.h
```diff
@@ -90,80 +90,79 @@ namespace PointMatcherSupport
...
 		{
 			classes[name] = descriptor;
 		}
-		
+
 		//! Return a descriptor following a name, throw an exception if name is invalid
-		const ClassDescriptor* getDescriptor(const std::string& name) const
+		std::shared_ptr<ClassDescriptor> getDescriptor(const std::string& name) const
```
YoshuaNava, line 162: This type is mentioned in line 156. Does it make sense to create an alias? That way we also abstract a bit the type of smart pointer and the library of origin.

pointmatcher/Registrar.h
```diff
@@ -174,42 +173,43 @@ namespace PointMatcherSupport
...
 		//! Create an instance
 		std::shared_ptr<Interface> create(const std::string& name, const Parametrizable::Parameters& params = Parametrizable::Parameters()) const
 		{
 			return getDescriptor(name)->createInstance(name, params);
 		}
-				
+
 		//! Create an instance from a YAML node
-        	std::shared_ptr<Interface> createFromYAML(const YAML::Node& module) const
+		std::shared_ptr<Interface> createFromYAML(const YAML::Node& module) const
```
YoshuaNava, line 184: Ditto alias.

pointmatcher/Registrar.h
AlexandreG87: Hi @YoshuaNava,
I'm sorry, but I'm not quite sure to understand what you mean here. Do you mean creating an alias for `std::shared_ptr<ClassDescriptor>` or replacing the `typedef` syntax for the `using` syntax?

@pomerlef and @simonpierredeschenes what do you think?

YoshuaNava: In here I mean replacing typedef with using.

In the later comments I do propose an alias for the shared pointer of the class descriptor.

pomerlef: To be frank, I never used `using` in C++ except in `using namespace ...`, but I'm not an expert neither. @YoshuaNava is most probably in a better position to answer that.

When I'm search for `using` in the library, it's clearly not uniform. A recent contribution in octree is using it, but that's about it. To not branch multiple PR in one, I would keep `typedef` here and open an issue about using `using` properly through the code.

YoshuaNava: @pomerlef Thanks. Agreed 👍

pointmatcher/Registrar.h
AlexandreG87: From what I understand, there's no method provided to modify the descriptor, so it seems that it never change, but I can be wrong. I'm not enough familiar with the library to be 100% sure. These changes were made with the help of @simonpierredeschenes.

For the const reference part, I've found [this](https://www.internalpointers.com/post/move-smart-pointers-and-out-functions-modern-c) to have the clearer explanation on how to pass a smart pointer to a function and I would tend to agree with you by passing it by const reference.

Thanks for your suggestion!

@pomerlef and @simonpierredeschenes What do you think?

pointmatcher/Registrar.h
AlexandreG87: Like I said, I'm not enough familiar with the library to tell what is the best thing to do. I've searched in the code if there was such an alias definition for a shared pointer and I haven't found any. What alias would you suggest for this type?

@pomerlef and @simonpierredeschenes What do you think?

pointmatcher/Registrar.h
AlexandreG87: Same as the previous comment.

pointmatcher/Registrar.h
PhiBabin: In our usecase, the reg function is only called in the registrar and in that usecase the smart pointer is an rvalue and we are transfering ownership. A const reference can not be used without creating an useless temporary variable. So there are two ways of doing this:
pass by value
```
{
 ...
  reg_by_value("foo", std::make_shared<Foo>()); // Copy constructor
....
}
void reg_by_value(const std::string &name, std::share_ptr<ClassDescriptor> descriptor)
{
  classes.insert(std::make_pair(name, std::move(descriptor))); // move construction
}
```
rvalue reference:
```
{
 ...
  reg_by_rvalue_ref("foo", std::make_shared<Foo>());
....
}
void reg_by_rvalue_ref(const std::string& name, std::share_ptr<ClassDescriptor>&& descriptor)
{
  classes.insert(std::make_pair(name, descriptor)); // move construction
}
```
The second solution prevent a copy. However, the interface now required a rvalue reference, which would be a problem if this function is used outside of the library.

YoshuaNava: Makes sense. Thank you :+1:

pointmatcher/Registrar.h
YoshuaNava: This item is a bit more of a nitpick, I'd say :slightly_smiling_face: 

Feel free to address it or discard it.

pointmatcher/Registrar.h
AlexandreG87: Thank you @PhiBabin, these are really interesting solutions. I've found in [this article](https://herbsutter.com/2013/06/05/gotw-91-solution-smart-pointer-parameters/) that passing a temporary shared pointer as parameter (which is the case here), it will be move construct. So, I think that it would make your first solution look like this?

```cpp
{
 ...
  reg_by_value("foo", std::make_shared<Foo>()); // move construction
 ...
}
void reg_by_value(const std::string &name, std::share_ptr<ClassDescriptor> descriptor)
{
  classes.insert(std::make_pair(name, descriptor));
}
```
Tell me what you think and feel free to correct me... I'm still learning the C++ subtleties! :)

pointmatcher/Registrar.h
AlexandreG87: No problems! I'll leave it like that then. Thanks!

pomerlef:
Are we ready to merge?

AlexandreG87:
Not yet, I was waiting for feedback on how we should pass the shared pointer in the reg method.

pointmatcher/Registrar.h
PhiBabin: You are right, in C++11, in this example when we use `reg_by_value` we will have two move constructions (not 1 copy + 1 move construction), where as `reg_by_rvalue_ref` will have a single move construction.

AlexandreG87: Thanks @PhiBabin, let's go for this then! 👍🏻

AlexandreG87:
@pomerlef I think we are good to merge now. 👍🏻

Issue #222, opened 2017-10-03
Is there a Python binding to libpointmatcher


s0r2637:
I was wondering if there is a Python binding to libpointmatcher?

pomerlef:
Clearly not, but I would highly appreciate it ;)

pomerlef:
We starting to work on Python binding. If someone has some inputs, advices or needs. Now is the time to tell us!

PhiBabin:
There was the start of an implementation of python bindings. It supported registration, add numpy array as pointcloud input and support extracting the debugging information that are normally in the .vtk output. However, it needs to be clean up, there is a bunch of code that was made specifically for debugging penalized ICP. https://github.com/norlab-ulaval/test_penalized_icp

YoshuaNava:
@PhiBabin How did you attempt the Python binding in your branch?

AlexandreG87:
@YoshuaNava Thanks for the informations! I'll take a look at them! 🙂

@PhiBabin I don't have access to the repo you've mentionned. 🙁

PhiBabin:
@YoshuaNava  I used boost.python.
@AlexandreG87 Oups, my bad, it is now public.

AlexandreG87:
Alright! So, after researching about all the Python bindings tools available, here are the ones I consider to have good potential to do the job. These solutions have been selected base on the following criteria :

### Primary criteria:

- Performance
- Modern C++ and Python3 compatibility
- Effective Numpy to Eigen conversion support (not 100% sure for cppyy)
- Template compatibility
- The ease of use and maintenance
- The solution must still be actively maintained
- No or minimal modification to the original code must be done

### Secondary criteria:

- Availability/Quality of documentation and tutorials
- Control and flexibility over interface implementation
- Does not require too much dependency

#### The fully automatic solutions :

- [**Automatic Wrapper and Interface Generator (AutoWIG)**](https://autowig.readthedocs.io/en/latest/) :
    - [Article on AutoWIG](https://peerj.com/articles/cs-149/)
    - [YouTube video explaning how AutoWIG works](https://youtu.be/N4q_Vud77Hw)
    - [GitHub repo](https://github.com/StatisKit/AutoWIG)

- [**cppyy**](https://cppyy.readthedocs.io/en/latest/) :
    - [Paper on cppyy](http://wlav.web.cern.ch/wlav/Cppyy_LavrijsenDutta_PyHPC16.pdf)
[...]

Issue #306, opened 2018-11-26
Remove boost dependencies


PhiBabin:
Removing dependencies such as Boost will make the inclusion of of libpm easier. This is the list of task for which we still use boost
- [ ] String formating: There is a lot of header only library that could be added to replace boost::format. There also more modern alternative such as [fmt](https://github.com/fmtlib/fmt). They could be easily added to contrib/.
- [ ] Filesystem: It's hard to have a multi platform alternative to boost::filesystem.[ Luckily c++17 now support part](https://en.cppreference.com/w/cpp/experimental/fs) of the boost api. gcc 8.0 is [the first version to support it](https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html#status.iso.2017). However, most of our user are still Ubuntu 16.04 and on Ubuntu 18.04, which have gcc 4.5 and 7.3 respectively. We could add a macro to cmake to switch between the standard library and boost depending on the version of gcc.

PR #397, opened 2020-08-06
[WIP] feature/python_module : Adding libpointmatcher's Python bindings (#222)

AlexandreG87:
Here are the python's bindings for libpointmatcher as requested in the issue #222.

#### Things to do before merging :

- [x] Make sure that the module get installed in the sites-packages of the current Python environment (maybe using pip install?).

- [x] Provide a tutorial on how to compile and install the module.

- [x] Provide a tutorial on how to use it.

- [x] Finish to transpose the C++ examples into Python.

- [x] Merging PR #395

- [ ] Make sure that the bindings compile on macOS?

#### Testing :

If you want to test the Python module on your side, clone my libpointmatcher's [fork](https://github.com/AlexandreG87/libpointmatcher). Then, follow the compilation and installation instructions from this [tutorial](https://github.com/AlexandreG87/libpointmatcher/blob/feature/python_module/doc/CompilationPython.md).

I've tested it with g++ 7.5.0 ~~and clang++ 10.0.0~~ on Ubuntu 18.04.

**Edit:** The bindings no longer compiles on clang++ 10.0, and on clang++ 6.0 either, since the adding of the IO module.

Feedback on the bindings themselves and the tutorials will be really appreciated, thanks! :slightly_smiling_face:

***Last edited :*** 20 august 2020


pointmatcher/IO.cpp
```diff
@@ -1780,9 +1780,9 @@ template
 class PointMatcherIO<double>::PLYElement;
 
 template
-class PointMatcherIO<float>::PLYProperty;
+struct PointMatcherIO<float>::PLYProperty;
```
YoshuaNava, line 1783: What is the reason behind this change?

python/pointmatcher/io.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+#include "pointmatcher/IO.h"
+
+PYBIND11_MAKE_OPAQUE(PointMatcherIO<ScalarType>::FileInfoVector)
+
+namespace pointmatcher
+{
+	void pybindIO(py::module& p_module)
+	{
+		using PMIO = PointMatcherIO<ScalarType>;
```
YoshuaNava, line 11: Does it make sense to make this alias available in the namespace?

python/pointmatcher/io.cpp
```diff
@@ -0,0 +1,217 @@
...
+		py::bind_vector<SupportedLabels>(pyPointMatcherIO, "SupportedLabels", "Vector of supported labels in PointMatcher and their external names");
+
+		using GenericInputHeader = PMIO::GenericInputHeader;
+		py::class_<GenericInputHeader>(pyPointMatcherIO, "GenericInputHeader", "Helper structure designed to parse file headers")
+			.def_readwrite("name", &GenericInputHeader::name)
+			.def_readwrite("matrixRowId", &GenericInputHeader::matrixRowId)
+			.def_readwrite("matrixType", &GenericInputHeader::matrixType)
+			.def(py::init<>())
+			.def(py::init<const std::string&>(), py::arg("name"));
+
```
YoshuaNava, line 43: Could you add comments describing what each of the above blocks do?

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.11)
+cmake_minimum_required(VERSION 2.8.12)
```
YoshuaNava, line 1: Why this particular version of cmake?

Btw, this change is not reflected in the changelog.

doc/CompilationPython.md
```diff
@@ -0,0 +1,189 @@
...
+
+| Name | Version <br> (Tested August 2020 on Ubuntu 18.04) |
+| :--- | :---: |
+| pybind11 | 2.5.0 |
+| Python3 | 3.6.9 |
+|||
+| **Dependencies** | |
+| catch | 1.10.0 |
+| python3-dev | 3.6.7 |
+| pytest | 5.4.3 |
```
YoshuaNava, line 20: Is this a test-only depend?

doc/CompilationPython.md
```diff
@@ -0,0 +1,189 @@
...
+
+```bash
+sudo apt install catch python3-dev
+```
+
+and `pytest` with `pip` :
+
+```bash
+pip3 install pytest
+```
```
YoshuaNava, line 32: The above instructions are only for Ubuntu. Could you specify that?

doc/CompilationPython.md
```diff
@@ -0,0 +1,189 @@
...
+
+> ***Note :*** pybind11 is also compatible with Python2.7, which means that it is also possible to generate a module for Python2.7. But, this hasn't been tested. You can refer to the pybind11 [official documentation](https://pybind11.readthedocs.io/en/latest/index.html) for more informations on what you can do with pybind11 and how to do it.
```
YoshuaNava, line 34: Python 2.7 is also EOL: https://www.python.org/doc/sunset-python-2/

doc/CompilationPython.md
```diff
@@ -0,0 +1,189 @@
...
+ 
+Once this is done, run the following commands :
+
+```bash
+mkdir build && cd build
+cmake ..
+make check -j 4
+```
+
+This will both compile and run the pybind11 tests. Next, you can install pybind11 by running this command :
```
YoshuaNava, line 61:
```suggestion
This will both compile and run the pybind11 tests. Next, you can install pybind11 by running this command:
```

doc/CompilationPython.md
```diff
@@ -0,0 +1,189 @@
...
+
+```bash
+sudo make install
```
YoshuaNava, line 64: Where does this install the bindings? Usual /usr/lib, or /home/<user>/.local/python3x/ ?

The second is the usual one for packages fetched via pip.

doc/CompilationPython.md
```diff
@@ -0,0 +1,189 @@
...
+```
+
+At the moment, pypointmatcher can only be compiled with **version 2.5.0** of pybind11. To install the right version, run the following commands :
+
+```bash
+cd pybind11
+git checkout v2.5.0
+```
+
+Before going any further, the `CMakeLists.txt` file from the `/python` directory needs to be modified as follows :
```
YoshuaNava, line 93: Could this instead be added to the CMake file with the use of conditionals and a variable to choose the pybind fetching method?

doc/CompilationPython.md
```diff
@@ -0,0 +1,189 @@
...
+
+> ***IMPORTANT :*** When this method is used, it is very important to checkout the version **2.5.0** of pybind11 or it will be impossible to generate the build files.
+
+Once this is done, return to libpointmatcher's `/build` directory.
+
+You're now ready to proceed to the [configuration step](#configuration).
+
+## Configuring the variables <a name="configuration"></a>
+
+> ***Note :*** *It is recommended to create a virtual python environment before proceeding with the next steps.*
```
YoshuaNava, line 128: Can you explain why is it recommended to use a virtual env?

doc/CompilationPython.md
```diff
@@ -0,0 +1,189 @@
...
+> ***Note :*** If you are using the system's python environment, replace the `site.getsitepackages()[0]` by `site.getusersitepackages()`.
+
+> ***IMPORTANT :*** *This last example is the default behavior if no path has been set before compiling the module.* ***Please, make sure that this correspond to a valid location or this will lead to an import error.***
+
+#### Enabling the compilation
+
+By default, pypointmatcher compilation is disabled. In order to compile it, you must set the CMake variable `BUILD_PYTHON_MODULE` to `ON` :
+ 
+```bash
+cmake -D BUILD_PYTHON_MODULE=ON ..
```
YoshuaNava, line 168: Nice 👍

examples/align_sequence.cpp
```diff
@@ -58,8 +58,8 @@ int main(int argc, char *argv[])
 {
 	validateArgs(argc, argv);
 
-	typedef PointMatcher<float> PM;
-	typedef PointMatcherIO<float> PMIO;
+	typedef PointMatcher<double> PM;
+	typedef PointMatcherIO<double> PMIO;
```
YoshuaNava, line 62: Does this change affect the performance in any way?

python/CMakeLists.txt
```diff
@@ -0,0 +1,118 @@
...
+	pointmatcher/io.cpp
+	pointmatcher/matcher.cpp
+	pointmatcher/matches.cpp
+	pointmatcher/outlierfilter.cpp
+	pointmatcher/outlierfilters.cpp
+	pointmatcher/pointmatcher.cpp
+	pointmatcher/transformation.cpp
+	pointmatcher/transformations.cpp
+	pointmatcher/transformationchecker.cpp
+	pointmatcher/transformationcheckers.cpp
```
YoshuaNava, line 36: Why the change in capitalization of source files?

python/CMakeLists.txt
```diff
@@ -0,0 +1,118 @@
...
+					${CMAKE_SOURCE_DIR}/pointmatcher/DataPointsFilters/utils
+					${CMAKE_SOURCE_DIR}/pointmatcher/ErrorMinimizers)
+
+if (pybind11_FOUND)
+	message(STATUS "pybind11 v${pybind11_VERSION}")
+
+	pybind11_add_module(pypointmatcher ${PYBIND11_SOURCES})
+
+	target_link_libraries(pypointmatcher
+						  PRIVATE
```
YoshuaNava, line 107: Is this file written using spaces or tabs? It is not displayed uniformly on my browser.

python/datapointsfilters/covariancesampling.cpp
```diff
@@ -0,0 +1,31 @@
+#include "covariancesampling.h"
```
YoshuaNava, line 1: Name of this file is not snake-cased, as promoted by Python.

python/pypointmatcher_helper.h
```diff
@@ -0,0 +1,72 @@
+#ifndef PYTHON_PYPOINTMATCHER_HELPER_H
+#define PYTHON_PYPOINTMATCHER_HELPER_H
+
```
YoshuaNava, line 3: The name of this file is not very clear. To me it is more like a centralized typedefs-header.

YoshuaNava (commented):
Hi @AlexandreG87,
This PR looks really cool. Just left a few minor comments.

pointmatcher/IO.cpp
AlexandreG87: I changed it because I was getting this warning in CLion :

`Class 'PLYProperty' was previously declared as a struct; this is valid, but may result in linker errors under the Microsoft C++ ABI`

doc/CompilationPython.md
AlexandreG87: For `catch` and `pytest` yes, not for `python3-dev`.

doc/CompilationPython.md
AlexandreG87: For the apt package manager yes, I could add the macOS command. But for the pip command, isn't it the same for macOS and Ubuntu?

doc/CompilationPython.md
AlexandreG87: I knew, but I put it there just in case some people would like to know that fact. I have no problem to remove it, if you think that it is irrelevant.

CMakeLists.txt
AlexandreG87: I choose this version because according to the issue #306, the user base of libpointmatcher were still on Ubuntu 16.04 almost two years ago. So, I assumed that nobody were using Ubuntu 12.04 anymore. Then, @simonpierredeschenes and I decided that it was probably safe enough to change it since cmake 2.8.12 is shipped with Ubuntu 14.04. That being said, if the majority of the users are using Ubuntu 16.04, I guess it would be safe to update the version to 3.5.1. But, I know that it will break the CMakeLists.txt and that would require some additional work to make it work again. Please, tell me if this makes sense, I can be wrong, I'm still learning that kind of stuff. 🙂

About the changelog, I guess it would be mentioned for the next release?

examples/align_sequence.cpp
AlexandreG87: I would say that this might affect the performances, but I don't know for sure. I would ask @simonpierredeschenes what are his thoughts about this.

Also, the reason these changes were made is that these examples were used as a reference to determine if the behaviour of the python examples matched that of the C++ examples. So, knowing that the results of the both versions were the same gave me a good indication that what I was doing was right.

python/CMakeLists.txt
AlexandreG87: These are the source files for the bindings only. Since it was my first time making python bindings for a C++ library, I found these [guidelines](https://developer.lsst.io/v/DM-8980/pybind11/style.html) and tired to follow some of them. But now, I realize that I misinterpreted [that guideline](https://developer.lsst.io/v/DM-8980/pybind11/style.html#wrappers-for-a-c-header-file-should-go-in-a-python-module-with-a-lowercased-version-of-the-header-file-name) that was suggesting to use lowercase module and not lowercase source files. I thought it was to minimize ambiguity between the real source files and the wrapper source files. I will eventually rename them with the right convention. 😅

python/CMakeLists.txt
AlexandreG87: I used tabs. I have the same issue in my browser too, but the display were fine in my IDE. I've changed my reformat settings for cmake files and it will be fix in a future commit. 👍🏻

python/datapointsfilters/covariancesampling.cpp
AlexandreG87: I wasn't sure which convention to use. I explained the reason behind this in my other comment. I'm fine using snake_case python convention.

python/pypointmatcher_helper.h
AlexandreG87: I agree that it ended up being that way, but in the main purpose of this file was to regroup all the necessary include in one place and also the `PYBIND11_MAKE_OPAQUE` macro declaration as mentioned in pybind11 [documentation](https://pybind11.readthedocs.io/en/latest/advanced/cast/stl.html#making-opaque-types). I thought about whether I put these aliases just before their use, like the io.cpp source file, or in a common header file like it is right now. Some of them are used across some other files, so I did this to avoid re-declaring them every time I needed it. But, if you suggest that it would be better to declare it only where it is needed (saying that make me realize that it makes much more sense to do so 😅), I have no problem to fix this.

doc/CompilationPython.md
AlexandreG87: the `python3-dev` package doesn't have to be installed on macOS, the [documention](https://pybind11.readthedocs.io/en/latest/basics.html#linux-macos) says that the included python version works out of the box. I can specify that instead.

doc/CompilationPython.md
AlexandreG87: This will install pybind11, not the bindings, in the `/usr/local/include` directory. It is up to the user to specify where the bindings will be installed. If no location are specified, the default path will be `/usr/local/lib/python3.x/dist-packages` for the system python environment, which is not recommended. Instead, it should be `/home/username/.local/lib/python3.x/site-packages` , like you said. And it would something like `/home/username/python_virtual_env_dir/virtual_env_name/lib/python3.x/site-packages` for a virtual python environment. I think this is well explained in the tutorial. I could add another warning about not using the `dist-packages` directory, since it is reserved for packages installed with the system package manager. At least for Linux system AFAIK.

AlexandreG87: I've checked where libnabo python bindings were installed in order to take a decision. I have the python2 and python3 versions installed on my computer. The python2 version is installed in the `/usr/local/lib/python2.7/dist-packages` directory and the python3 version is installed in the `/usr/local/lib/python3.6/dist-packages` and in the `/home/username/.local/lib/python3.6/site-packages` directory. So, what would be the best option?

doc/CompilationPython.md
AlexandreG87: Yes, it can. It could be a cmake variable `USE_INSTALLED_PYBIND11` or `USE_SYSTEM_PYBIND11` that would be set to `ON` be default.

doc/CompilationPython.md
AlexandreG87: Yes, I can or I thought that I could put a link to a tutorial that would explain how and why creating a virtual environment. I think that [The Coding Train](https://youtu.be/nnhjvHYRsmM) tutorial (youtube) and [Real Python](https://realpython.com/python-virtual-environments-a-primer) tutorial are good ones. What do you think?

python/pointmatcher/io.cpp
AlexandreG87: These aliases are intended to be use only locally and to make the next declaration shorter and clearer, instead of always using something like `PointMatcherIO<ScalarType>::PLYElement::methodName` for each method and member of the class. But, I'm always open to suggestions. 🙂

python/pointmatcher/io.cpp
AlexandreG87: Yes, I will eventually add comments to describe the whole thing. I also thought maybe I could add a section to the tutorials with a link to the pybind11 documentation for those who would like to contribute to the bindings.

AlexandreG87:
@YoshuaNava Thank you for the feedback and I hope you will be satisfied with my answers. 😄

pointmatcher/IO.cpp
AlexandreG87: Thanks! 🙂

Now, to adress your questions :

1. I use CLion's built-in inspections, as well as clang-tidy via the built-in clangd version 11.
2. No, only under Ubuntu 18.04.

python/pointmatcher/io.cpp
AlexandreG87: Probably in this PR, but it can go in another one as well.

doc/CompilationPython.md
AlexandreG87: pybind11's tests are built by default and require `pytest` and `catch`. But, if `catch` is not installed, it will tried to fetch it and install it, so this one is kind of optional.

YoshuaNava:
@AlexandreG87 What is the status of this PR?

* I can give this a try under Windows. Do you also plan to test the bindings generation when compiling with ROS?

AlexandreG87:
> @AlexandreG87 What is the status of this PR?

Hi @YoshuaNava, I would say that it's not quite finished yet. I'll start to address your comments first, then we'll see what needs to be done to make it better.

>     * I can give this a try under Windows. Do you also plan to test the bindings generation when compiling with ROS?

It was not in my plans, I'll need to check with @pomerlef what can be done about that.

AlexandreG87:
> Before approving I have some final questions:
> 
>     1. On which platforms have you tested the bindings?
> 
>     2. Which versions of Python are you supporting? Is there any performance variation or change needed to run with Python 2/3?
> 
>     3. Have you tested both building/running, but also installing the bindings? If so, where do they get installed?
> 
>     4. Does the uninstall target also remove the bindings?

1. So far, I've only been testing on Ubuntu 18.04 with Python 3.6.9 and pybind11 2.5.0.

1. We will support Python 3 only, since Python 2 is deprecated.

1. Yes, I've tested building and running, but also installing. You can find these answers in the compilation tutorial.

1. Yes, it does.

python/datapointsfilters/covariancesampling.cpp
AlexandreG87: All files have been renamed according to the Python convention. So, I'll mark this as resolved. Thanks!

python/pypointmatcher_helper.h
AlexandreG87: Alright, I'll see what I can do about that.

examples/align_sequence.cpp
AlexandreG87: After a few quick tests, the float version is actually slightly faster than the double version, both for C++ and Python libpointmatcher version. So, I'll change the compile type of the bindings to float and revert the type for the examples as well.

CMakeLists.txt
AlexandreG87: > Btw, is there any official position on which version of Ubuntu we support?

Not really, I guess? But it would probably be a good idea to have one.

YoshuaNava: Maybe this question is a bit out of scope. It makes sense to open a separate issue and resolve this thread.

doc/CompilationPython.md
YoshuaNava: I see. Np.

examples/align_sequence.cpp
YoshuaNava: Oki. You can still change it easily and re-export it, right?

Can that detail about the Python binding be included in the documentation?

YoshuaNava:
@AlexandreG87 Regarding your answers:

> Thanks, it's very appreciated. Maybe, I'll give it a try too.

Oki.

> It was not in my plans, I'll need to check with @pomerlef what can be done about that.

I would say that for minimum support we should make sure that the ROS build still runs. In a separate PR I plan to submit a new CMake setup, that allows you to have ROS CI/CD, running tests, and at the same time build without catkin, in case wanted. I'll open an issue to propose it. 

> 1. So far, I've only been testing on Ubuntu 18.04 with Python 3.6.9 and pybind11 2.5.0.
> 2. We will support Python 3 only, since Python 2 is deprecated.

That's good :+1: 

image

> 3. Yes, I've tested building and running, but also installing. You can find these answers in the compilation tutorial.
> 4. Yes, it does.

Very good. No more comments from my side apart from the open ones.

examples/align_sequence.cpp
AlexandreG87: > Oki. You can still change it easily and re-export it, right?

Sorry, I'm not sure I'm following you on this one. Do you mean changing the type of the bindings? If so, yes I just need to the type of my alias for the PointMatcher instance.

> Can that detail about the Python binding be included in the documentation?

Yes, I can and just to be sure, the detail you're talking about is the fact that the bindings are compiled using float, right?

YoshuaNava: I meant with my question if it's possible to change the bindings to work with float with minimal overhead.

Yes, mentioning that the work with floats :)

pomerlef:
I would go on an merge this PR and open issues if something go wrong. For now, I don't have time to test it on MacOs.

AlexandreG87:
> I would go on an merge this PR and open issues if something go wrong. For now, I don't have time to test it on MacOs.

@pomerlef There were some more things that I wanted to do before you merge this PR and I will do them as soon as possible.

AlexandreG87:
@pomerlef You can proceed to the merge when you're ready.

PR #399, opened 2020-08-07
Devel/nimisbert

nimisbert:
Hello,

I added 5 robust weight functions. They are the weight functions of Sparse ICP, Fair, Logistic, Andrew, EM-ICP.


pomerlef:
ok to test

PR #382, opened 2020-06-01
Modernize cmake; make cmake compatible with git submoduling

jlblancoc:
This PR does:
* Use `GNUInstallDirs` to correctly install files in their standard directories (multiarch-compatible).
* Add `install(EXPORT ...` to cmake scripts to allow better use of exported targets.
* Add build options to allow not building parts of the projects (defaults=as it was before)
* Automatic detection of `libnabo` if built as part of the same source tree (e.g. git submodules).

PR #406, opened 2020-09-23
Compatibility with libnabo

iabaldwin:
With the recent CMake change in libnabo (abd2e94), the previous
linking/include directives are incorrect. This change works with libnabo
built from source, and installed via `.deb`.


pomerlef:
Could you have a look at the PR #382? We are trying to resolve this problem with libnabo as soon as possible over there.

iabaldwin:
Apologies, didn't realize it was in progress. Will close this PR out.

PR #411, opened 2020-10-19
[feature/spdf] Add SpectralDecompositionFilter (SpDF)

MathLabu:
Add a new DataPointsFilter, namely the Spectral Decomposition Filter based on the following reference:
> Labussière, M., Laconte, J., & Pomerleau, F. (2020). Geometry Preserving Sampling Method Based on Spectral Decomposition for Large-Scale Environments. Frontiers in Robotics and AI, 7(September), 1–15. https://doi.org/10.3389/frobt.2020.572054

------------------------------
Spectral Decomposition Filter (SpDF) is a sampling algorithm based on spectral decomposition analysis to derive local density measures for each geometric primitive.
 * First, we identify the geometric primitives along with their saliencies using the tensor voting framework. 
 * Then, we derive density measures from saliencies: if the density for each geometric primitive is less than the desired density, we stop; else we sub-sample each over-represented geometric primitive, and re-iterate.
 * As output, we have a uniform sampled point cloud enhanced with geometric information.

I added the implementation of the Tensor Voting framework based on the closed-form solution of 
> Wu, T.-P., Yeung, S.-K., Jia, J., Tang, C.-K., & Medioni, G. (2012). A Closed-Form Solution to Tensor Voting: Theory and Applications. IEEE Transactions on Pattern Analysis and Machine Intelligence, 34(8), 1482–1495. https://doi.org/10.1109/TPAMI.2011.250

-------------------------
if the PR is accepted, the branches `feature/tensorvoting`, `feature/spdf` and `feature/spdf-experimental`are no longer required and can be deleted.


pomerlef (approved):

pomerlef:
I only passed through it rapidly. Good job for the comments and reference. Do you want to add a quick entry in [DataFilters.md](https://github.com/ethz-asl/libpointmatcher/blob/master/doc/DataFilters.md) to make sure people finds it?

MathLabu:
Thanks! DataPointsFilters documentation have been updated by adding entries for the `Spectral Decomposition Filter (SpDF)` and the `Saliency Filter`.

PR #417, opened 2020-10-29
Link against pthread

YoshuaNava:
I added [pthread](https://computing.llnl.gov/tutorials/pthreads/) to the list of external libraries that libpointmatcher links against.

This usually fixes pthread linking issues for other packages. Do note that I haven't been able to reproduce this on my PC.

@peci1 @lucaBartolomei You reported linking issues with pthread before, can you verify that this PR solves your issue?


lucaBartolomei:
Thanks for taking care of this! After this change, the linking problems against phtread disappears.

YoshuaNava:
Great!

Issue #421, opened 2020-11-25
Windows Platform: linker is complaining about possibly missing static library pthread.lib 


princetn:
Severity	Code	Description	Project	File	Line	Suppression State
Error	LNK1104	cannot open file 'pthread.lib'	build_map	D:\GIT\libpointmatcher\build\examples\LINK	1	
Error	LNK1104	cannot open file 'pthread.lib'	align_sequence	D:\GIT\libpointmatcher\build\examples\LINK	1	
Error	LNK1104	cannot open file 'pthread.lib'	icp_customized	D:\GIT\libpointmatcher\build\examples\LINK	1	
Error	LNK1104	cannot open file 'pthread.lib'	list_modules	D:\GIT\libpointmatcher\build\examples\LINK	1	
Error	LNK1104	cannot open file 'pthread.lib'	icp_simple	D:\GIT\libpointmatcher\build\examples\LINK	1	
Error	LNK1104	cannot open file 'pthread.lib'	pmicp	D:\GIT\libpointmatcher\build\examples\LINK	1	
Error	LNK1104	cannot open file 'pthread.lib'	icp_advance_api	D:\GIT\libpointmatcher\build\examples\LINK	1	
Error	LNK1104	cannot open file 'pthread.lib'	compute_overlap	D:\GIT\libpointmatcher\build\examples\LINK	1

princetn:
To fix this, I had to download pthreadVC2.lib for the x64 version from this link: https://sourceware.org/pthreads-win32/ 
you can get to the ftp link: ftp://sourceware.org/pub/pthreads-win32/dll-latest/lib/ and choose which version you works for you.
Next, you then have to change under the linker for all the projects with the issue the path where you downloaded this static library.

PR #428, opened 2020-12-10
Search for pthread or pthreadVC2 lib in CMakeLists

maximecharriere:
Following [issue 421](https://github.com/ethz-asl/libpointmatcher/issues/421)

pthread is a POSIX library. So Windows can't compile project that include `pthread.h`.
Windows use threads included in `windows.h`.
To use pthread, Windows users have to get [pthreads-win32](https://sourceware.org/pthreads-win32/).
The library provided by pthreads-win32 is called `pthreadVC2.lib`.
So CMake have to search for pthread and pthreadVC2

I'm new to GitHub and CMake, tell me if I'm doing something wrong.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -114,7 +114,8 @@ add_subdirectory(contrib)
 #---------------------------
 # DEPENDENCY:  POSIX Threads
 #---------------------------
-set(EXTERNAL_LIBS ${EXTERNAL_LIBS} pthread)
+find_library(pthread_LIB_DIR  NAMES pthread pthreadVC2)
+set(EXTERNAL_LIBS ${EXTERNAL_LIBS} ${pthread_LIB_DIR})
 
 #--------------------
 # DEPENDENCY:  boost


pomerlef:
ok to test

pomerlef:
@peci1 will it break something on your side?

peci1:
Looking through the repo, I'm not even sure pthread is really needed. I only see usages of boost::thread. But I'm not sure if boost pulls in pthread automatically or not. But definitely, boost::thread should not rely on pthread on windows.

The standard way of getting pthread is `find_package(Threads)` and then linking against `Threads::Threads` imported target. This should correctly handle all operating systems and pull in their respective threading dependencies. But the imported target is only available since CMake 3.1. As libpointmatcher tries to support 2.8, it should instead use variable `CMAKE_THREAD_LIBS_INIT` when linking targets that require threading support.

So, I'd suggests something like

    find_package(Threads REQUIRED)
    if (CMAKE_VERSION VERSION_LESS 3.1.0)
      set(EXTERNAL_LIBS ${EXTERNAL_LIBS} ${CMAKE_THREAD_LIBS_INIT})
    else()
      set(EXTERNAL_LIBS ${EXTERNAL_LIBS} Threads::Threads)
    endif()

pomerlef:
Is that PR still valid or #437 has priority on that one?

maximecharriere:
#437 is a nicer solution because the user doesn't have to download pthread

PR #436, opened 2021-01-26
Fix windows doc

maxGimeno:
Replace the name of the libnabo cmake variables to match the ones in the cmake scripts
(like [here](https://github.com/ethz-asl/libpointmatcher/blob/master/CMakeLists.txt#L367))


pomerlef:
ok to test

pomerlef:
Thanks a lot for adapting the documentation !

PR #437, opened 2021-01-26
Fixes for windows

### LINKED ISSUE #435: Problem with boost libraries in libpointmatcherConfig.cmake
Hello,

I'm trying to use PointMatcher 1.3.1 on Windows with MSVC2019, because a dependency to pthread in master make it impossible to use on Windows.

After building pointmatcher, I call find_package(libpointmatcher) in my project, and I get cmake errors when trying to generate the project:
```
CMake Error at CMakeLists.txt:96 (add_executable):
  Target "registration_with_pointmatcher" links to target "Boost::chrono" but
  the target was not found.  Perhaps a find_package() call is missing for an
  IMPORTED target, or an ALIAS target is missing?
 ```
and one for each boost component required.
I looked a little and I found out that the configuration file `libpointmatcherConfig.cmake` adds the boost components to the list of external libraries, as it should, but it is missing a `find_package` call.
If I manually add `find_package(Boost COMPONENTS thread filesystem system program_options date_time chrono)` in libpointmatcherConfig.cmake, I don't get any cmake error anymore.

maxGimeno:
- Fixes the pthread dependency as discussed in #428 
 - Fixes the boost dependency declaration in libpointmatcherConfig.cmake.in so the users won't have to declare it in their project.
 
I also have a bug, that says that `yaml-cpp-pm-targets.cmake` cannot be found, and I have to manually copy it from `/build/contrib/yaml-cpp-pm/CMakeFiles/Export/share/libpointmatcher/cmake/` to `./build` but I couldn't find how to fix it. 

Fixes #435

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -114,7 +114,12 @@ add_subdirectory(contrib)
 #---------------------------
 # DEPENDENCY:  POSIX Threads
 #---------------------------
-set(EXTERNAL_LIBS ${EXTERNAL_LIBS} pthread)
+find_package(Threads REQUIRED)
+if (CMAKE_VERSION VERSION_LESS 3.1.0)
+  set(EXTERNAL_LIBS ${EXTERNAL_LIBS} ${CMAKE_THREAD_LIBS_INIT})
+else()
+  set(EXTERNAL_LIBS ${EXTERNAL_LIBS} Threads::Threads)
+endif()
 
 #--------------------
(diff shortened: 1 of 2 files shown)


pomerlef:
ok to test

pomerlef:
Thanks for the PR!

@peci1 anything breaking on your side before I merge?

peci1:
Thanks for asking, this looks harmless ;)

PR #439, opened 2021-01-30
Add option to weight by reading pc

hermannsblum:
This extends the feature to weight by the point intensity in the map.


pomerlef (approved):

pomerlef:
Thanks for the PR!

PR #441, opened 2021-02-03
Add Boost_LIBRARIES to pointmatcher_LIBRARIES

maxGimeno:
This follows #437 . I forgot to add the Boost_LIBRARIES to the lists of pointmatcher_LIBRARIES, so when linking with libpointmatcher, they are still misisng.


pomerlef:
ok to test

PR #446, opened 2021-02-11
Add libpointmatcher_INCLUDE_DIR to match with CGAL

maximecharriere:
When using libpointmatcher with CGAL, `<pointmatcher/PointMatcher.h>` can't be found because libpointmatcher is not added to the list of included directories.  
The reason for this is that CGAL wait for the `libpointmatcher_INCLUDE_DIR` variable when libpointmatcher set the `libpointmatcher_INCLUDE_DIRS` variable (with a `S`)
  
-----
[CGAL_pointmatcher_support.cmake](https://github.com/CGAL/cgal/blob/master/Installation/cmake/modules/CGAL_pointmatcher_support.cmake)

```cmake
if(libpointmatcher_FOUND AND NOT TARGET CGAL::pointmatcher_support)
  add_library(CGAL::pointmatcher_support INTERFACE IMPORTED)
  set_target_properties(CGAL::pointmatcher_support PROPERTIES
    INTERFACE_COMPILE_DEFINITIONS "CGAL_LINKED_WITH_POINTMATCHER"
    INTERFACE_INCLUDE_DIRECTORIES "${libpointmatcher_INCLUDE_DIR}"
    INTERFACE_LINK_LIBRARIES "${libpointmatcher_LIBRARIES}")

endif()
```


pomerlef:
ok to test

pomerlef:
thanks!

PR #448, opened 2021-02-12
Automaticaly find eigen3

maximecharriere:
Now we have to set the CMake variable `EIGEN_INCLUDE_DIR` manually when installing libpointmatcher on Windows.  
It would be simpler to use the environment variable `EIGEN3_INC_DIR` added when installing Eigen.  

To make this changes, I was inspired by the [FindEigen3.cmake](https://github.com/CGAL/cgal/blob/master/Installation/cmake/modules/FindEigen3.cmake) file from CGAL.  
I didn't take the whole file which is very complete and secure, but only the part that find Eigen path.  

I was not able to test these modifications on Linux. So it should be done before merge.


pomerlef:
ok to test

pomerlef:
thanks again @maximecharriere !

PR #453, opened 2021-03-15
[Matches/OutlierFiltersImpl] Made convergence error messages more informative

YoshuaNava:
This PR aims to make the convergence error messages from the Matches and OutlierFilters classes a bit more informative.


pomerlef:
Thanks!

PR #464, opened 2021-05-31
Fix weird behavior of MaxPointCountDataPointsFilter

### LINKED ISSUE #463: Getting weird results/behavior from MaxPointCountDataPointsFilter
As the title suggest, I'm getting weird results from `MaxPointCountDataPointsFilter` as the point clouds don't seem to really have the number of point specified by the `maxCount` parameter.

Here's a videos to illustrate what I'm talking about:

[Current results video](https://youtu.be/yw0PLw_2KJE)

I will open a PR to show the results I get with the fix I made.

aguenette:
fixes #463

Like mentioned in the issues #463, this is the fix that I'm proposing in order to get better and more consistent results with the filter.

Here's a video showing the results:

[Fixed results video](https://youtu.be/m9oe-D1ZHY4)

### DIFF
--- a/pointmatcher/DataPointsFilters/MaxPointCount.cpp
+++ b/pointmatcher/DataPointsFilters/MaxPointCount.cpp
@@ -33,20 +33,21 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 */
 #include "MaxPointCount.h"
+#include <random>
 
 // MaxPointCountDataPointsFilter
 // Constructor
 template<typename T>
 MaxPointCountDataPointsFilter<T>::MaxPointCountDataPointsFilter(const Parameters& params):
-	PointMatcher<T>::DataPointsFilter("MaxPointCountDataPointsFilter", 
+	PointMatcher<T>::DataPointsFilter("MaxPointCountDataPointsFilter",
 		MaxPointCountDataPointsFilter::availableParameters(), params),
 	maxCount(Parametrizable::get<size_t>("maxCount"))
 {
-	try 
+	try
 	{
 		seed = this->template get<size_t>("seed");
-	} 
-	catch (const InvalidParameter&) 
+	}
+	catch (const InvalidParameter&)
 	{
 		seed = static_cast<size_t>(1); // rand default seed number
 	}
@@ -54,7 +55,7 @@ MaxPointCountDataPointsFilter<T>::MaxPointCountDataPointsFilter(const Parameters
 
 // Compute
 template<typename T>
-typename PointMatcher<T>::DataPoints 
+typename PointMatcher<T>::DataPoints
 MaxPointCountDataPointsFilter<T>::filter(const DataPoints& input)
 {
 	DataPoints output(input);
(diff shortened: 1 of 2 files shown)


pointmatcher/DataPointsFilters/MaxPointCount.cpp
```diff
@@ -66,35 +67,20 @@ MaxPointCountDataPointsFilter<T>::filter(const DataPoints& input)
 template<typename T>
 void MaxPointCountDataPointsFilter<T>::inPlaceFilter(DataPoints& cloud)
 {
-	const size_t N = static_cast<size_t> (cloud.features.cols() - 1);
-	
-	if (maxCount <= N) 
+	const auto N = static_cast<size_t>(cloud.features.cols() - 1);
```
YoshuaNava, line 70:
```suggestion
	const size_t N{static_cast<size_t>(cloud.features.cols() - 1)};
```

If you use a known type with bracket initialization you can benefit from the compiler informing you in case there are narrowing casts 
https://stackoverflow.com/questions/42975620/in-c-can-we-use-for-c-style-casting

pointmatcher/DataPointsFilters/MaxPointCount.cpp
```diff
@@ -66,35 +67,20 @@ MaxPointCountDataPointsFilter<T>::filter(const DataPoints& input)
...
+
+	if (maxCount <= N)
 	{
-		//Re-init seed at each call, to ensure same results
-		std::srand(seed);
-		
-		for(size_t j=0; j<maxCount; ++j)
+		std::default_random_engine randomNumberGenerator(seed);
+
+		for(size_t j = 0; j < maxCount; ++j)
```
YoshuaNava, line 76:
```suggestion
		for(size_t j{0u}; j < maxCount; ++j)
```

pointmatcher/DataPointsFilters/MaxPointCount.cpp
```diff
@@ -66,35 +67,20 @@ MaxPointCountDataPointsFilter<T>::filter(const DataPoints& input)
...
-				cloud.descriptors.col(j) = cloud.descriptors.col(idx);
-				cloud.descriptors.col(idx) = desc;
-			}
-			if (cloud.times.cols() > 0)
-			{
-				const auto time = cloud.times.col(j);
-				cloud.times.col(j) = cloud.times.col(idx);
-				cloud.times.col(idx) = time;
-			}	
+			std::uniform_int_distribution<size_t> distribution(j, N);
```
YoshuaNava, line 79: How expensive is this operation?

pointmatcher/DataPointsFilters/MaxPointCount.cpp
```diff
@@ -66,35 +67,20 @@ MaxPointCountDataPointsFilter<T>::filter(const DataPoints& input)
...
-	if (maxCount <= N) 
+	const auto N = static_cast<size_t>(cloud.features.cols() - 1);
+
+	if (maxCount <= N)
 	{
-		//Re-init seed at each call, to ensure same results
-		std::srand(seed);
-		
-		for(size_t j=0; j<maxCount; ++j)
+		std::default_random_engine randomNumberGenerator(seed);
```
YoshuaNava, line 74: What is the default random engine? Did you evaluate it against other options?

pointmatcher/DataPointsFilters/MaxPointCount.cpp
```diff
@@ -66,35 +67,20 @@ MaxPointCountDataPointsFilter<T>::filter(const DataPoints& input)
...
-			{
-				const auto time = cloud.times.col(j);
-				cloud.times.col(j) = cloud.times.col(idx);
-				cloud.times.col(idx) = time;
-			}	
+			std::uniform_int_distribution<size_t> distribution(j, N);
+			const size_t index = distribution(randomNumberGenerator);
+
+			//Switch columns j and index
+			cloud.swapCols(j, index);
```
YoshuaNava, line 83: Can it happen that you swap a point that has already been swapped?

pointmatcher/DataPointsFilters/MaxPointCount.cpp
aguenette: In the [documentation](https://en.cppreference.com/w/cpp/numeric/random), it says that the `default_random_engine` is implementation-defined and on my computer (`gcc` 7.5.0) it says that it's `std::minstd_rand0`.

I did some benchmark to evaluate the different ways random numbers can be generated:

int_vs_real_minstd_rand0

int_vs_real_minstd_rand

real_minimal_standard_vs_mersenne_twister_double

real_minimal_standard_vs_mersenne_twister_float

From that, we can clearly see that only modifying the old algorithm to use the C++11 `random` library is faster than using C `rand` function, and it also produce better results at shuffling `DataPoints`.

pointmatcher/DataPointsFilters/MaxPointCount.cpp
aguenette: I think this is a possibility, but I don't know if it can be a problem. To avoid this, we would need keep track of the points that has been already swapped. Do you think it can be a problem?

pomerlef:
Don't hesitate to open new issues if needed. I'll assume that this PR is ready for merge.

Issue #376, opened 2020-04-23
Organized point clouds


YoshuaNava:
(This issue continues the discussion started in #342 about organized point clouds)

Organized point clouds are an attractive way to represent 3D point sets coming from range sensors. The main advantage of them is that they are akin to matrices, thus easier to traverse. Furthermore, the neighboring set of each point is known a priori, so operations like normal computations could potentially skip the K-NN step, reducing computation time and CPU/memory taken in building a K-D tree.

Existing implementations of organized point clouds (pcl, sensor_msgs/PointCloud2) represent the range measurements in a flat data structure akin to a matrix, where each cell corresponds to a range sensor bin. Thus the range sensor FOV is discretized horizontally and vertically. Missing points are usually filled with NaN.

I'm opening this issue to discuss how organized point clouds could be represented within the framework of libpointmatcher. Some points for discussion I can think of right now:

* Whether we can re-use/extend the current DataPoints class to host the missing point ordering information.
* If it would make sense to actually use a different representation. For example, a matrix with multiple channels, one for each descriptor (similar to a cv::Mat from OpenCV)

YoshuaNava:
I'm currently working on speeding up surface normals computation on DataPoints structures. I'm planning to implement an 'Index Image', a structure that contains a mapping between a 2D matrix (an organized point cloud grid) and 1D (DataPoints) elements. This approach is inspired by [Serafin et al](http://jacoposerafin.com/nicp/)

I consider it a better option than depth images / depth buffers, as it seems more general and straightforward to perform a mapping in 3D space with a direction table, than to map point clouds from arbitrary sensors into depth images (for example, dome LIDARs).

We plan to use this as a mean to speed up feature sampling filters, initially.

PR #465, opened 2021-06-17
Fix the surface normals datapoints filter covariance matrix bug

aguenette:
There was a bug in the computation of the covariance matrix, which had an impact in the eigen values and eigen vectors computation.

In the `SurfaceNormals` filter: the covariance matrix needed to be divided by `realKnn`.
In the `SamplingSurfaceNormals` filter: the covariance matrix needed to be divided by `colCount`.

Credits go to @pomerlef.

### DIFF
--- a/pointmatcher/DataPointsFilters/SamplingSurfaceNormal.cpp
+++ b/pointmatcher/DataPointsFilters/SamplingSurfaceNormal.cpp
@@ -245,7 +245,7 @@ void SamplingSurfaceNormalDataPointsFilter<T>::fuseRange(
 	const Matrix NN = (d.colwise() - mean);
 
 	// compute covariance
-	const Matrix C(NN * NN.transpose());
+	const Matrix C((NN * NN.transpose()) / T(colCount));
 	Vector eigenVa = Vector::Identity(featDim-1, 1);
 	Matrix eigenVe = Matrix::Identity(featDim-1, featDim-1);
 	// Ensure that the matrix is suited for eigenvalues calculation
--- a/pointmatcher/DataPointsFilters/SurfaceNormal.cpp
+++ b/pointmatcher/DataPointsFilters/SurfaceNormal.cpp
@@ -184,7 +184,7 @@ void SurfaceNormalDataPointsFilter<T>::inPlaceFilter(
 		const Vector mean = d.rowwise().sum() / T(realKnn);
 		const Matrix NN = d.colwise() - mean;
 
-		const Matrix C(NN * NN.transpose());
+		const Matrix C((NN * NN.transpose()) / T(realKnn));
 		Vector eigenVa = Vector::Zero(featDim-1, 1);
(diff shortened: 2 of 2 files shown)


pomerlef:
add to whitelist

pomerlef (approved):

YoshuaNava:
@aguenette so the computed eigenvalues had an unexpected scale?

pomerlef:
yes exactly. We pass from
wrong
to
dense_mapper_markers2



@YoshuaNava do you know who handle Jenkins at ASL? They changed something and the CI pipeline is broken again.

aguenette:
@pomerlef I need to rectified something about the pictures you posted. The first one was taken using the `SamplingSurfaceNormals` filters using 30 `knn`, a `ratio` of 0.5 (I think? 🤔) and the square root of the eigen values was not taken (after that it is). The second one used the `SurfaceNormals` filters with 15 `knn` (this one I'm sure 😉).

I made a mistake by assuming that both filters were giving the same results. So, after seeing the question of @YoshuaNava yesterday, I ran some tests with both filters before and after the correction were applied.

So, here's some pictures taken from those tests:

`SamplingSurfaceNormals` before:

SamplingSurfaceNormals_before

and after:

SamplingSurfaceNormals_after

`SurfaceNormals` before:

SurfaceNormals_before

and after:

SurfaceNormals_after

15 `knn` was used for both filters and a ratio of 1 was used for the `SamplingSurfaceNormals` filter.

So, it's looks like that the `SamplingSurfaceNormals` was worst than the `SurfaceNormals` filter and that it's still problematic even after applying the correction.

Sorry about that. 😓

YoshuaNava:
@pomerlef Sorry for the delay, I was focused on a software release sprint. At the moment I'm not aware of who is handling the CI from ASL.

@aguenette Thank you for the very detailed explanation :slightly_smiling_face:. I have two questions that are closely related:

1. Have you observed any change in the performance (accuracy, robustness, snappiness) of ICP after applying the fix?
2. Which tool do you use for plotting normals? This question is relevant for --> https://github.com/ethz-asl/libpointmatcher/issues/376#issuecomment-870630306

aguenette:
> @aguenette Thank you for the very detailed explanation slightly_smiling_face. I have two questions that are closely related:
> 
>     1. Have you observed any change in the performance (accuracy, robustness, snappiness) of ICP after applying the fix?
>     2. Which tool do you use for plotting normals? This question is relevant for --> [Organized point clouds #376 (comment)](https://github.com/ethz-asl/libpointmatcher/issues/376#issuecomment-870630306)

1. This fix shouldn't affect the performance of ICP because it only affect the scaling of the eigen values, i.e. the eigen vectors stay the same, and only the smallest eigen value determine which eigen vector is the normal.

2. The above pictures were taken using `rviz` `Marker` display type with a `MarkerArray` message. (https://wiki.ros.org/rviz/DisplayTypes/Marker)

PR #466, opened 2021-06-22
[BIV-1934] Obstacle detection using norlab_icp_mapper

JMtSG:
Added getter function for match distances to map


JMtSG:
Mistakenly requested

PR #467, opened 2021-07-01
Apply the transformation to eigen vectors

aguenette:
This PR adds the eigen vectors to the descriptors that need to be transformed in the `RigidTransformation` and `SimilarityTransformation` methods, because it was not the case when they were present in the descriptors. This was causing issues when trying to visualize them and working with them.

### DIFF
--- a/pointmatcher/TransformationsImpl.cpp
+++ b/pointmatcher/TransformationsImpl.cpp
@@ -65,30 +65,46 @@ void TransformationsImpl<T>::RigidTransformation::inPlaceCompute(
 	assert(cloud.features.rows() == parameters.rows());
 	assert(parameters.rows() == parameters.cols());
 
-	const unsigned int nbRows = parameters.rows()-1;
-	const unsigned int nbCols = parameters.cols()-1;
+	if(this->checkParameters(parameters) == false)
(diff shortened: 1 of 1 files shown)


YoshuaNava:
@aguenette The PR looks good to me. Nonetheless, I wanted to ask if it would be possible to create in a follow-up a centralized definition of commonly used point cloud fields. For example, normals, eigVectors, densities, etc? Such that any user of these fields can fetch the definitions from there.

aguenette:
> @aguenette The PR looks good to me. Nonetheless, I wanted to ask if it would be possible to create in a follow-up a centralized definition of commonly used point cloud fields. For example, normals, eigVectors, densities, etc? Such that any user of these fields can fetch the definitions from there.

Hi @YoshuaNava! Thanks for your review. About the centralized definition of the point cloud fields, the [`DataPointsFilters` page](https://github.com/ethz-asl/libpointmatcher/blob/master/doc/DataFilters.md) already provide a short definition of the different fields present in the filters, but some filters miss some fields. So, it could be a good start to add the missing ones and improve their definitions or maybe we could add a section at the end of the `DataPointsFilters` page who would provide a more detailed definition of them. What do you think? Were you talking about something like this?

PR #470, opened 2021-07-09
Windows: Fix Narrowing conversion of seed in MaxPointCount

### LINKED ISSUE #469: Narrowing conversion on windows
I get 
```
libpointmatcher\pointmatcher\DataPointsFilters\MaxPointCount.cpp(73,42): error C2398: Element '1': conversion from 'size_t' to 'unsigned int' requires a  a narrowing conversion
```
When building with MSVC 2019 on Windows.

maxGimeno:
Fix #469

### DIFF
--- a/pointmatcher/DataPointsFilters/MaxPointCount.cpp
+++ b/pointmatcher/DataPointsFilters/MaxPointCount.cpp
@@ -70,7 +70,7 @@ void MaxPointCountDataPointsFilter<T>::inPlaceFilter(DataPoints& cloud)
 
 	if (maxCount <= N)
 	{
-		std::minstd_rand randomNumberGenerator{seed};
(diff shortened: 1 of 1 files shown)


lrineau:
The constructor of `std::minstd_rand` expects a `std::uint_fast32_t` as first argument.
  - On a x86_64 Linux machine, that type is an `unsigned long int`, and thus a 64 bits unsigned.
  - It seems that on Windows with the compiler from MSVC, that type is `unsigned int`, and thus only a 32 bits integer.

And thus on Windows, passing `seed` to that `std::uint_fast32_t` argument is a narrowing conversion. As the constructor was called with `{seed}` and not `(seed)`, that narrowing conversion triggered an error.

pomerlef:
Ok to test

pomerlef:
Thanks!

PR #482, opened 2021-12-09
catkin not required for pure cmake packages

Timple:
This prevens rosdep resolving in ROS2


pomerlef:
ok to test

pomerlef:
@simonpierredeschenes can you check if this patch will break default ROS1 installs ?

simonpierredeschenes:
I'll have a look

Timple:
For what it's worth: it has beendone before for pure cmake packages :slightly_smiling_face: 
https://github.com/nobleo/paho.mqtt.c-release/blob/master/patches/package.xml

PR #490, opened 2022-03-03
Use the LOG_INFO_STREAM macro instead of std::cerr

maximecharriere:


maximecharriere:
Does not work. Wait until I update the PR.

pomerlef:
ok to test

pomerlef:
thanks!

PR #495, opened 2022-05-16
Add more details in pypointmatcher's installation docs

IamPhytan:
* Add commands for systems with multiple versions of Python
    * Explicitly specify that we use Python 3.6 when installing `pytest` and when building `pybind11`
* Code block for creating a venv when appropriate


IamPhytan:
Yes

PR #501, opened 2022-10-12
Add seed to Random sampling filter

boxanm:
I added the seed parameter, already present for other filters, such as Maximum point count or NSS. I think that eventually this parameter should be added to all filters that contain randomness to allow a fair comparison between e.g. sampling methods.


pomerlef:
add to whitelist

YoshuaNava:
Thank you!!!

YoshuaNava:
Btw, for follow-up. In those unit tests that rely on randomness, we could record the random seed, such that we can still 'test the filters randomly' (random is, ultimately, their intended behavior), but still be able to debug test failures.

See http://google.github.io/googletest/reference/testing.html#Test::RecordProperty

pomerlef:
ok to test

pomerlef:
Took me a while to pin point the error in the log:
```
fatal: unable to connect to github.com:
github.com[0: 140.82.121.4]: errno=Connection timed out
```
The ASL CI server is not behaving properly and I'm not sure who is maintaining that anymore.

PR #506, opened 2023-01-03
Build python binding as regular package

KernelA:
This pull request contains:

- GitHub workflow to build python package for Linux only
- publishing it on the release page when tag a commit
- new doc for python binding

[Release page example](https://github.com/phygitalism/libpointmatcher/releases/tag/3.0.1test)

We can install python binding as regular python package via pip. I think it simplifies package using and will lead to more broadly using.

The workflow contains code for windows building, but I cannot compile it with MS MPI. Changes in the source code are needed.


pomerlef:
@aguenette can you have a look?

pomerlef:
@KernelA , this is a great add-on for libpointmatcher! 

We are not set up to use GitHub workflow yet, but the timing is good as our legacy CI tool seems to stop working. Hopefully, by mid-summer, we will have transitioned to GitHub workflow. Don't give up, your PR will be processed ;)

KernelA:
@pomerlef, ok. I think there is some work. For example, extracting version from CMake file as base version for the python package.  We must sync it manually, at the moment.

RedLeader962:
I'm on it

RedLeader962:
@KernelA By the way awesome work. It is a great addition to libpointmatcher.

python/setup.py
```diff
@@ -0,0 +1,29 @@
...
+    """Distribution which always forces a binary package with platform name
+    """
+    def has_ext_modules(self):
+        return True
+
+path_to_file = "../pointmatcher/PointMatcher.h"
+
+version = None
+
+with open(path_to_file, "r", encoding="utf-8") as f:
```
RedLeader962, line 14: Fetching the pointmatcher version dynamicaly will be a very usefull tool for the build system automation.
(!) Limitation: This approach to fetching the `POINTMATCHER_VERSION` only work if the user execute

``` bash
python -m build --wheel
```

It will break when using the default

``` bash
python -m build
```

because the build command will build *sdist* in a first step and then build *wheel* fetching from a temporary dir `/tmp/build-via-sdist-*/pypointmatcher-*`.
I recommand
- to keep this approach for now but wrapped in a try/except clause with an explicit error message pinpointing its origin since it is not obvious when it happen in a install stacktrace which is usualy prety long;
- to add an explicit comment on those limitations in the code to advise futur developer;
- and to add a warning in the `doc/CompilationPython.md` package install procedure.

python/setup.py
```diff
@@ -0,0 +1,29 @@
+from setuptools import setup, find_packages
+from setuptools.dist import Distribution
+
+class BinaryDistribution(Distribution):
+    """Distribution which always forces a binary package with platform name
+    """
+    def has_ext_modules(self):
+        return True
+
+path_to_file = "../pointmatcher/PointMatcher.h"
```
RedLeader962, line 10: (!) Hardcoding path break python portability (to non unix OS in that case). Use [`os.path` package](https://docs.python.org/3/library/os.path.html) to define path instead, it will handle OS specific filesystem difference.
``` suggestion
path_to_file = os.path.relpath(os.path.join("pointmatcher", "PointMatcher.h"))
```

RedLeader962 (changes requested):
I've finish my review and added my recommandation. Go job @KernelA this is going to be very useful for the build system automation. 
I've tested the GitHub workflow using [nektos/act](https://github.com/nektos/act).
Note to self: we will have to refactor the jobs steps eventualy to reduce build time (boost and libnabo install could go in a docker container) if we want to add support for other ubuntu versions and computer architecture but this pull request is a great starting point.
Thanks @KernelA

KernelA:
The last build status:
https://github.com/phygitalism/libpointmatcher/actions/runs/4966179029

PR #513, opened 2023-05-31
Handle libnabo config mode

sloriot:
An alternative would be to set libnabo_INCLUDE_DIRS and libnabo_LIBRARIES in `libnabo-targets.cmake`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -162,13 +162,17 @@ include_directories(SYSTEM "${EIGEN_INCLUDE_DIR}")
 if (NOT TARGET nabo)
   # Find libnabo:
   find_package(libnabo REQUIRED PATHS ${LIBNABO_INSTALL_DIR})
-  message(STATUS "libnabo found, version ${libnabo_VERSION} (include=${libnabo_INCLUDE_DIRS} libs=${libnabo_LIBRARIES})")
+  if (TARGET libnabo::nabo)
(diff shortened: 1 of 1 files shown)


pomerlef:
ok to test

pomerlef:
@sloriot I'm missing the intent of the PR. Can you describe what was not working or what you want to improve?

sloriot:
I wanted to have the [config mode of cmake](https://cmake.org/cmake/help/latest/command/find_package.html) working with libnabo. Indeed, if you use the config mode then libnabo will be reported as found but the cmake variables that pointmatcher expects are not set.

As a side note, I also noticed that it is not possible to use pointmatcher with cmake without doing `make install` which does not seems to be ideal for people without root access to their computer. Last point, I don't understand why libnabo is not part of the dependency libs of pointmatcher target like you did with boost. I hacked it with https://github.com/CGAL/cgal/pull/7498 but if you are interested in getting some cmake update I can submit another PR to address both issues.

pomerlef:
Any link with #514? and of course, if you want to propose another PR we will gladly review it!

sloriot:
> Any link with #514? and of course, if you want to propose another PR we will gladly review it!

It indeed seems to be the same issue that my hack fixes at CGAL level. Since there was no such error in the past, I guess that some libnabo API is now exposed in pointmatcher (thus the need for the link).

pomerlef:
Thanks @sloriot !

PR #535, opened 2023-11-21
New DataPoints filter for descriptor augmentation

boxanm:
This DataPointFilter adds a new descriptor with a given **name**, **dimension**, and **values** to the point cloud on which it is applied. It can be used, for example, when a map-building algorithm requires the scans to have the same descriptor as the existing map.

### DIFF
--- a/pointmatcher/DataPointsFilters/AddDescriptor.cpp
+++ b/pointmatcher/DataPointsFilters/AddDescriptor.cpp
@@ -27,14 +27,32 @@ typename PointMatcher<T>::DataPoints AddDescriptorDataPointsFilter<T>::filter(
 // In-place filter
 template<typename T>
 void AddDescriptorDataPointsFilter<T>::inPlaceFilter(
-	DataPoints& cloud)
+        DataPoints& cloud)
 {
     Matrix matrix = PM::Matrix::Ones(descriptorDimension, cloud.getNbPoints());
     for(std::size_t i = 0; i < descriptorDimension; ++i)
     {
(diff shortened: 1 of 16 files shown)


boxanm:
I'm putting a **TODO** here to implement Python bindings for the filter ❌

pointmatcher/DataPointsFilters/AddDescriptor.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+// In-place filter
+template<typename T>
+void AddDescriptorDataPointsFilter<T>::inPlaceFilter(
+        DataPoints& cloud)
+{
+    Matrix matrix = PM::Matrix::Ones(descriptorDimension, cloud.getNbPoints());
+    for(std::size_t i = 0; i < descriptorDimension; ++i)
+    {
+        matrix.row(i) *= descriptorValues[i];
```
simonpierredeschenes, line 35: why is it a multiplicating instead of overwriting

simonpierredeschenes (changes requested):
Good job, there is only one small comment to address

pointmatcher/DataPointsFilters/AddDescriptor.cpp
boxanm: The Matrix `matrix` is initialized as Ones and each row is multiplicated by the corresponding descriptor value. Another option would be to assign `descriptorValues` to each column on `matrix`. Idk what's more optimal

Issue #454, opened 2021-03-19
[OctreeGridDataPointsFilter] Random points get discarded / added during execution, specially at high resolutions


YoshuaNava:
We use libpointmatcher's Octree Grid filter for a point cloud-based SLAM system with positive results. It is a really cool filter :+1:

However, we have noticed that the octree grid filter seems to discard or add points when resampling the octree up to a certain density level. This happens after long-term operation with low density grids (approx. 10cm), but becomes more pronounced as you increase the resolution of the octree, up to the point where, at resolutions of 2-4cm, the octree seems to add new points, and discard chunks of points in certain regions.

When it comes to points getting discarded, the phenomenon seems to manifest in 3 different ways:

1. The whole point cloud loses density.
2. Parts of the map get round or square-shaped holes.
3. Crevasses suddenly appear in the point clouds. (see image below, red highlight)

---

In the following figures you can observe some examples of points getting discarded. Notice the square area in the top right part of this map (perspective view):

![Screenshot_2021-03-19_11-19-15](https://user-images.githubusercontent.com/7463211/111765880-2c59dc00-88a5-11eb-94ef-6bd796c4fb56.png)

Orthogonal perspective of the same square hole, which seems to be getting new points, but rather slowly, with respect to the rest of the map:

![Screenshot_2021-03-19_11-18-18](https://user-images.githubusercontent.com/7463211/111766041-5ad7b700-88a5-11eb-84cb-6ae6d9a78e4a.png)

After more time has passed, notice the square hole in the top right stays, along with three horizontal 'crevasses' (one of them right next to the white circle):

![image](https://user-images.githubusercontent.com/7463211/111765819-1cda9300-88a5-11eb-9402-924908e6dfe7.png)

We have not collected screenshots from the case of new points getting added to the point clouds, but it is something that can be observed by monitoring the points around a sensor, while it stays static, or checking the extremas of a map, e.g. which can increase in z even if no points are integrated from the sensor with a strong z component.

---

We are very interested in fixing this issue, as the filter seems to have really good performance, and fulfills our requirements. I'm opening this issue with the intention of reporting it for others, and getting ideas on how to debug, and address it properly.

(FYI @MathLabu)

pomerlef:
Thanks for letting us know! That's indeed troublesome as we are using it also in some core applications. Indeed, if @MathLabu have an inside, it would be great.

MathLabu:
[...]

PR #540, opened 2023-12-04
Fix octree sampling

boxanm:
I accidentally noticed this bug when unit testing an NDT-inspired octree sampling method.
The current octree sampling implementation takes into account the situation when the indexed point was already processed; see this example:
```C++
std::size_t j = d;
//retrieve index from lookup table if sampling in already switched element
if(std::size_t(d)<idx)
   j = mapidx[d];
```
However, we don't know whether `mapidx[d]` doesn't again point to an index that was also processed. What actually needs to be done, is to propagate all the changes that were made on all previously used indexes of `mapidx`. So in this PR, I replaced the above code example and similar by
```C++
std::size_t j = d;
//retrieve index from lookup table if sampling in already switched element
while(j < idx)
   j = mapidx[j];
```
I haven't had time to check other Octree implementations and see how they deal with this situation.
Naturally, the use of the while cycle slows down the execution. There's probably a more clever solution using pointers?


simonpierredeschenes:
@boxanm Could you please  that this is the same issue as #454?

boxanm:
The bug I described shouldn't make random parts of the cloud disappear. Instead, the bug causes incorrect sampling inside individual cells. So points, that should be sampled are preserved and the output contains more points, than the number of cells in the octree. Therefore I don't think #454 is related.

PR #567, opened 2024-05-07
Set minimum libnabo version in CMakeLists.txt

boxanm:
I removed the minimum `libnabo` version that used to be set in `MatchersImpl.h` and moved it to `CMakeLists.txt` instead.


boxanm:
The compiling tests are expected to fail because `libnabo` version used to be set in CMakelists.txt with a regex reading a header file that hasn't been updated since 1.0.7 (see this [line](https://github.com/norlab-ulaval/libnabo/blob/68a0dc9c463c6b7c6cb20e84879f1f4e673f1dc7/CMakeLists.txt#L15)). We first need to merge [this PR](https://github.com/norlab-ulaval/libnabo/pull/137) which reads `package.xml` instead.

simonpierredeschenes (approved):

aguenette (dismissed):

aguenette (dismissed):

PR #588, opened 2024-10-14
Add support for binary PLY files

boxanm:
# Description

### Summary:
This PR adds support for binary PLY files, both loading and exporting.
The main limitation is that `libpointmatcher` treats all possible data types as either floats or doubles.

### Changes and type of changes (quick overview):

- The implementation is far from optimal. We could eventually switch to an external dependency, such as those mentioned in [this benchmark](https://github.com/mhalber/ply_io_benchmark?tab=readme-ov-file#libraries).
- Problems may arise with type overflows when loading PLY files containing `uint`, `int`, or `double` fields since they might not fit into `float`. However, such files are relatively rare, and the current implementation suffices for most use cases, for example, the [3DMatch dataset](https://3dmatch.cs.princeton.edu/).

---

# Checklist:

### Code related

- [x] I have made corresponding changes to the documentation 
      (i.e.: function, class, script header, README.md)
- [x] I have commented hard-to-understand code
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] All tests pass locally with my changes 
      (Check [contributing_instructions.md](/doc/contributing/contributing_instructions.md) for local testing procedure using _libpointmatcher-build-system_)

### PR creation related

- [x] My pull request `base ref` branch is set to the `develop` branch 
     (the _build-system_ won't be triggered otherwise)
- [x] My pull request branch is up-to-date with the `develop` branch 
     (the _build-system_ will reject it otherwise)

### PR description related

- [x] I have included a quick summary of the changes
- [x] I have indicated the related issue's id with `# <issue-id>` if changes are of type `fix`
- [x] I have included a high-level list of changes and their corresponding types
      (See [commit_msg_reference.md](/doc/contributing/commit_msg_reference.md) for details)

---


pointmatcher/IO.cpp
```diff
@@ -1573,21 +1574,94 @@ typename PointMatcherIO<T>::DataPoints PointMatcherIO<T>::loadPLY(std::istream&
...
+            throw runtime_error(
+                    (boost::format("PLY parse error: expected %1% values (%2% points with %3% properties) but only found %4% values.") % nbValues % nbPoints % nbProp %
+                     i).str());
+        }
+        else
+        {
+            const int row = vertex->properties[propID].pmRowID;
+            const PMPropTypes type = vertex->properties[propID].pmType;
+
+            // rescale color from [0,254] to [0, 1[
```
simonpierredeschenes, line 1658: The highest value in an 8-bit unsigned integer is 255, why does it end at 254?

simonpierredeschenes (approved):

pointmatcher/IO.cpp
boxanm: I think it's just a typo since the actual scaling uses 255.

lingerer:
There's possilbe problem for binary ply file reading.
 typename PointMatcherIO<T>::DataPoints PointMatcherIO<T>::loadPLY(const std::string& fileName)
this method use ifstream to read a ply file and don't specify using binary mode will fail to read some binary ply for unexpected file end.
possible solution should pre read ply header to verify if it's a binary ply or ascii ply.

another thought is this code should apply to binary pcd too.

boxanm:
Hey @lingerer, I'm not sure I understand - the decision on whether a .ply file is binary or not is made when processing the ifstream header part that was merged in this PR; See this line: https://github.com/norlab-ulaval/libpointmatcher/blob/3eb448330b2c35f7718e690e66f0c1057115e877/pointmatcher/IO.cpp#L1378
As for .pcd, `libpointmatcher` currently only supports its ASCII data format.
If you have other questions, consider creating a new issue. Thanks!

lingerer:
The same IO.CPP LINE 1284:

> ifstream ifs(fileName.c_str());

When I read a big ply file,I got a unexpected eof.
reading binary ply should use
ifstream ifs(fileName.c_str(),, std::ios::binary); to fix the potential problem.

And maybe that's why all libpointmatcher io type  support ascii type only

PR #607, opened 2025-07-25
Add timestamps to save ply

TomMontagnon:
# Description

### Problem

When I run:
```shell
ros2 service call /save_trajectory norlab_icp_mapper_ros/srv/SaveTrajectory "{trajectory_file_name: {data: '$PWD/my_traj.ply'}}"
```

It saves a trajectory, but without timestamps for each point.
Timestamps are passed to libpointmatcher, but the lib doesn't save them in case of `.ply`.
However, I need the timestamps, and I believe this is necessary in many cases.

(I was told lately that exporting to `.vtk` may save the timestamps, but I find working with `.ply` files to be more convenient anyway.)

### Short summary of the solution

Timestamps are passed as int64 to libpointmatcher, but the [PLY standard](https://gamma.cs.unc.edu/POWERPLANT/papers/ply.pdf) does not support int64.  
I therefore had to add two properties int32 to the PLY format: one for seconds and the other for nanoseconds (following the ROS2 convention).

# Main changes

- In `savePLY`, if the data contains times :
  - add two new properties "timestamp_sec" and "timestamp_nsec"
  - split the int64 into two int32 fields
- In `loadPLY`:
  - In phase three  `// 3- RESERVE DATAPOINTS MEMORY` : If "timestamp_sec" (int32) and "timestamp_nsec" (int32) are present, only allocate one int64 field initialized to 0
  - In phase four `// 4- PARSE PLY DATA (vertex)` : If I detect "timestamp_sec" (int32) and "timestamp_nsec" (int32), combine them with the appropriate coefficients into the previously-initialized int64 field.
- Added two new types "TIME" in `IO.h`
- Added a new unit test

# Checklist

- [x] The build passes
- [x] All original unit tests pass
- [x] I added a dedicated unit test for the "timestamps" case
- [x] I tested the branch locally for my use case: trajectory saving with timestamps works perfectly

# Other info

- [Line 80 in norlab_icp_mapper/Trajectory.cpp](https://github.com/norlab-ulaval/norlab_icp_mapper/blob/master/norlab_icp_mapper/Trajectory.cpp) shows that timestamps are passed correctly to libpointmatcher.
- [Line 64 in norlab_icp_mapper/Trajectory.cpp](https://github.com/norlab-ulaval/norlab_icp_mapper/blob/master/norlab_icp_mapper/Trajectory.cpp) shows that times are `int64_t`.
- Tracing the code back, [line 74 in norlab_icp_mapper_ros/mapper_node.cpp](https://github.com/norlab-ulaval/norlab_icp_mapper_ros/blob/ros2/src/mapper_node.cpp) indicates that these times come from `cloudMsgIn.header.stamp`, which based on the [ROS documentation](https://docs.ros.org/en/rolling/p/builtin_interfaces/msg/Time.html) is:
    ```
    int32 sec
    uint32 nanosec
    ```

### DIFF
--- a/pointmatcher/IO.cpp
+++ b/pointmatcher/IO.cpp
@@ -1559,12 +1559,29 @@ typename PointMatcherIO<T>::DataPoints PointMatcherIO<T>::loadPLY(std::istream& 
 
 	const unsigned int featDim = featLabelGen.getLabels().totalDim();
 	const unsigned int descDim = descLabelGen.getLabels().totalDim();
-	const unsigned int timeDim = timeLabelGen.getLabels().totalDim();
+	unsigned int timeDim_tmp = timeLabelGen.getLabels().totalDim();
+
+	bool found_sec = false, found_nsec = false, found_time = false;
+	for (const auto& lbl : timeLabelGen.getLabels())
+	{
+	    if (lbl.text == "timestamp_sec")
+	        found_sec = true;
+	    if (lbl.text == "timestamp_nsec")
+	        found_nsec = true;
+	    if (lbl.text == "time")
+	        found_time = true;
+	}
+
+	if (found_sec && found_nsec && !found_time)
+		--timeDim_tmp;
+	const unsigned int& timeDim = timeDim_tmp;
+
+		
 	const unsigned int nbPoints = vertex->num;
 
 	Matrix features = Matrix(featDim, nbPoints);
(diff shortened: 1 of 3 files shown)


boxanm:
Hey @TomMontagnon!
Thanks and good job on the PR! We've been quite busy on our site lately, but I will review the PR once things settle down a bit on our side.

PR #613, opened 2025-10-28
New DataPointFilter : Voxel HashMap 

nicolaslauzon:
# Description

### Summary:
This filter is inspired by [kiss-icp](https://github.com/PRBonn/kiss-icp/blob/main/cpp/kiss_icp/core/VoxelHashMap.cpp). It utilizes an HashMap to store the points in a voxel pattern. We limit the insertion to a maximum number of points per voxel which then allow for some filtering. Thanks to the HashMap, this filter is much faster than the VoxelGridDataPointFilter and is around two times slower than the RandomDownsamplingDataPointFilter.

### Filter in action:
image
The original point cloud is colored in blue. The subsampled point cloud is represented by the black points.

### Changes and type of changes (quick overview):

- Feature : New DataPointFilter named VoxelHashMap
- Feature : Python binding for this new filter
- Feature : Unit test to validate the new filter
- Docs : Update the python bindings compilation doc
- Docs : Documented the new VoxelHashMapDataPointFilter

---


doc/DataFilters.md
boxanm, line 1: Mention in the docs which points are kept (first in the array / random...)

pointmatcher/DataPointsFiltersImpl.h
boxanm, line 1: Can you revert the auto-formatting of this file?

pointmatcher/DataPointsFilters/VoxelHashMap.cpp
boxanm, line 1: From the code, I'm not sure the points are chosen randomly as mentioned in the header file

boxanm (commented):
Good job! I added some comments asking mostly for clarification in the docs.

nicolaslauzon:
> Good job! I added some comments asking mostly for clarification in the docs.

@boxanm thanks for the comments! I clarified the doc and reverted the autoformatting.

